PR #6, opened 2023-10-02
Add a test to ensure upgrades work

kfox1111:


faisal-memon:
Thanks @kfox1111 or this submission.

.github/workflows/helm-chart-ci.yaml
```diff
@@ -239,3 +239,44 @@ jobs:
 
       - name: Install and test example
         run: ${{ matrix.example }}/run-tests.sh
+
+  upgrade-test:
+    runs-on: ubuntu-22.04
```
faisal-memon, line 244: Do we need a stub in helm-chart-ci-ignore?

kfox1111: Good question. Not sure. Those are only for blocking tests? Should we make this blocking too or wait  for a bit to see how well it works?

faisal-memon: Good point. Not going to make this blocking for now.

examples/production/run-tests.sh
```diff
@@ -2,6 +2,9 @@
 
 set -xe
 
+UPGRADE_VERSION=v0.13.0
+UPGRADE_REPO=https://spiffe.github.io/helm-charts
```
faisal-memon, line 6:
```suggestion
UPGRADE_REPO=https://spiffe.github.io/helm-charts-hardened
```

examples/production/run-tests.sh
```diff
@@ -2,6 +2,9 @@
 
 set -xe
 
+UPGRADE_VERSION=v0.13.0
```
faisal-memon, line 5: Is this intended to be the latest version of the chart?

examples/production/run-tests.sh
kfox1111: Good catch

examples/production/run-tests.sh
kfox1111: ultimately I think? Was thinking we'd bump it on release for now, and work on an auto detection script that would set it to whatever is latest automatically in a future pr?

faisal-memon: haha read my mind with the auto detection.

faisal-memon (approved):

faisal-memon:
@kfox1111 Just need dco and we are good to go.

kfox1111:
Fixed.

I think we lost the option to auto signoff on merged suggestions.

faisal-memon:
> Fixed.
> 
> I think we lost the option to auto signoff on merged suggestions.

Yes, will fix later today.

faisal-memon (approved):

PR #7, opened 2023-10-02
Fix issue with version checker not running

faisal-memon:
The version updater workflows stopped working at [run #23](https://github.com/spiffe/helm-charts/actions/runs/5923605140)

Errror: `Restore cache failed: Dependencies file is not found in /home/runner/work/helm-charts/helm-charts. Supported file pattern: go.sum`

Open ticket with setup-go and this was what they said to do: https://github.com/actions/setup-go/issues/427#issuecomment-1731969319

### DIFF
--- a/.github/workflows/check-versions.yaml
+++ b/.github/workflows/check-versions.yaml
@@ -36,6 +36,7 @@ jobs:
         uses: actions/setup-go@v4.1.0
         with:
           go-version: '1.21'
+          cache: false
 
       - name: Setup crane
         uses: imjasonh/setup-crane@v0.3


kfox1111 (approved):
LGTM

.github/workflows/check-versions.yaml
```diff
@@ -36,6 +36,7 @@ jobs:
         uses: actions/setup-go@v4.1.0
         with:
           go-version: '1.21'
+          cache: false
```
marcofranssen, line 39:
```suggestion
          cache-dependency-path: ./tests
```

This would fix it for real, and still keep the caching to speed up builds.

The setup-go action looks for go.sum in the root of the repo by default. In our case the `go.mod` and `go.sum` files are located in the `tests` folder.

marcofranssen (changes requested):
See inline feedback.

PR #16, opened 2023-10-04
Add Tornjak ingress

mrsabath:


kfox1111 (commented):
LGTM.

Maybe we should have an example like:
https://github.com/spiffe/helm-charts-hardened/blob/main/examples/production/values-export-federation-https-spiffe-ingress-nginx.yaml
and:
https://github.com/spiffe/helm-charts-hardened/blob/main/examples/production/values-export-federation-https-web-ingress-nginx.yaml

where we show an example of how to do nginx support with passthrough of tls when authenticating via certs? Maybe one for openshift if that is different?

faisal-memon:
Thanks @mrsabath for introducing this change as a separate PR. Its more work for the submitter but helps the review process.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -562,6 +562,31 @@ tornjak:
...
+  ## @param tornjak.ingress.hosts [array] Host paths for ingress Tornjak backend service
+  ingress:
+    enabled: false
+    className: ""
+    annotations: {}
+    servicePort: 10000
+      # kubernetes.io/ingress.class: nginx
+      # kubernetes.io/tls-acme: "true"
+      # nginx.ingress.kubernetes.io/ssl-redirect: "true"
+      # nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
```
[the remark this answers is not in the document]
faisal-memon: If its annotations I think you need to move it up one line.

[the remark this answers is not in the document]
mrsabath: I think we don't need this right now. I am removing it

mrsabath:
https://github.com/spiffe/helm-charts-hardened/pull/16#pullrequestreview-1658365176 
Thanks @kfox1111 I was thinking about providing some examples and updated to README when we are done with OpenShift, as it is easier to demonstrate the steps in there.

faisal-memon (approved):

charts/spire/charts/spire-server/values.yaml
```diff
@@ -562,6 +562,27 @@ tornjak:
...
+  ## @param tornjak.ingress.enabled Flag to enable ingress for Tornjak backend service
+  ## @param tornjak.ingress.className Ingress class name for Tornjak backend service
+  ## @param tornjak.ingress.annotations [object] Annotations for Tornjak backend service
+  ## @param tornjak.ingress.servicePort Service port value for Tornjak backend service
+  ## @param tornjak.ingress.hosts [array] Host paths for ingress Tornjak backend service
+  ingress:
+    enabled: false
+    className: ""
+    annotations: {}
+    servicePort: 10000
```
marcofranssen, line 574:
```suggestion
```

This duplicates the port number across the service config and the ingress, which requires consumers to keep it aligned. It is better to use the port name in the ingress as defined in the service, see my other comment.

charts/spire/charts/spire-server/templates/tornjak-ingress.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: {{ $fullName }}
+  namespace: {{ include "spire-server.namespace" . }}
+  labels:
+    {{ include "spire-server.labels" . | nindent 4}}
+  {{- with .Values.tornjak.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{ include "spire-lib.ingress-spec" (dict "ingress" .Values.tornjak.ingress "svcName" (include "spire-tornjak.servicename" .) "port" .Values.tornjak.ingress.servicePort "global" .Values.global) | nindent 2 }}
```
marcofranssen, line 15:
```suggestion
  {{ include "spire-lib.ingress-spec" (dict "ingress" .Values.tornjak.ingress "svcName" (include "spire-tornjak.servicename" .) "port" "tornjak-srv-http" .Values.global) | nindent 2 }}
```

This uses the service port by name as defined in the Service definition.

marcofranssen (changes requested):
One small suggestion on using the servicePort by name so consumers don't have to align 2 values.

charts/spire/charts/spire-server/templates/tornjak-ingress.yaml
kfox1111: "global" got accidentally dropped in the suggestion?

marcofranssen: Forgot to remove the other global part, as there isn't a value required with this suggestion. Thanks for noticing my mistake.

marcofranssen (approved):
LGTM 🚀

Issue #5, opened 2023-10-02
Add support for custom plugins to SPIRE agent.


faisal-memon:
Server already has this feature, needs to be ported to the agent.

PR #22, opened 2023-10-06
Add plugin support to the spire agent

kfox1111:
This adapts the existing spire server plugin support to be usable by the agent as well.

fixes: https://github.com/spiffe/helm-charts-hardened/issues/5


charts/spire/charts/spire-agent/values.yaml
```diff
@@ -215,3 +217,20 @@ extraContainers: []
...
+# Upgrades wont be tested for anything under this config. If you need this, please let the chart developers know your needs so we
+# can prioritize proper support.
+## @skip unsupportedBuiltInPlugins
+unsupportedBuiltInPlugins:
+  keyManager: {}
+  nodeAttestor: {}
+  svidStore: {}
+  workloadAttestor: {}
+
+## @skip customPlugins
```
faisal-memon, line 231: I know we did the skip for server so it's fine to leave it as is. Just wondering if it would be good to document?

examples/spire-plugins/node-agent-tpm/values.yaml
```diff
@@ -23,3 +23,15 @@ spire-server:
         plugin_data:
         - ca_path: /run/spire/data/certs
           hash_path: /run/spire/data/hashes
+
+spire-agent:
+  image:
+    registry: docker.io
+    repository: kfox1111/misc
```
faisal-memon, line 30: Should we use an official image here?

faisal-memon:
Thanks @kfox1111 for filling in this gap.

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -24,26 +42,28 @@ agent:
...
 
   WorkloadAttestor:
-    - k8s:
-        plugin_data:
-          # Defaults to the secure kubelet port by default.
-          # Minikube does not have a cert in the cluster CA bundle that
-          # can authenticate the kubelet cert, so skip validation.
-          skip_kubelet_verification: {{ .Values.workloadAttestors.k8s.skipKubeletVerification }}
-          disable_container_selectors: {{ .Values.workloadAttestors.k8s.disableContainerSelectors }}
+  {{- if .Values.workloadAttestors.k8s.enabled }}
```
marcofranssen, line 54: Is there ever a situation where you don't want to do k8s attestors in a k8s deployment?

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -165,6 +165,8 @@ workloadAttestors:
     ## @param workloadAttestors.unix.enabled Enables the Unix workload attestor
     enabled: false
   k8s:
+    ## @param workloadAttestors.k8s.enabled Enables the Kubernetes workload attestor
+    enabled: true
```
marcofranssen, line 169: See previous comment. Is there ever a situation where you don't want k8s attestation in a Kubernetes deployment?

marcofranssen (commented):
A question on the new enabled flag see inline.

charts/spire/charts/spire-agent/templates/configmap.yaml
kfox1111: I'd like to use it for authenticating a nested spire server with the upstream-agent using just a tpm for extra security. So, I'd turn off the k8s one as it only supports the k8s node attestor

examples/spire-plugins/node-agent-tpm/values.yaml
kfox1111: There isn't one. Its the official image + the tpm plugin like in the Dockerfile. We could remove it and just put a placeholder there too. <your agent image here> ?

kfox1111: removed it in favor of something the user has to specify

faisal-memon:
Changes look good to me. @marcofranssen would appreciate a final review from you before merging.

marcofranssen (approved):
LGTM 🚀

PR #28, opened 2023-10-10
Fix failing Tornjak ingress port

mrsabath:
Deployment fails with error:

```
helm.go:84: [debug] error validating "": error validating data: ValidationError(Ingress.spec.rules[0].http.paths[0].backend.service.port.number): invalid type for io.k8s.api.networking.v1.ServiceBackendPort.number: got "string", expected "integer"
```

Setting a correct value for the Tornjak ingress port


faisal-memon (approved):

marcofranssen (approved):

kfox1111 (commented):
I reviewed this but forgot to hit submit. :(

There is another issue

PR #30, opened 2023-10-10
Add Tornjak ingress example

mrsabath:


examples/tornjak/values-ingress.yaml
```diff
@@ -0,0 +1,64 @@
+global:
+  spire:
+    jwtIssuer: https://oidc-discovery.$APP_SUBDOMAIN
```
kfox1111, line 3: I know what your trying to do and why, but it feels really wrong to try and template out values passed to a templating engine used for templating k8s objects.

There a feature we could add to handle some of this better? Been kind of on my todo list for a while. Maybe an globals.app_subdomain or something?

examples/tornjak/values-ingress.yaml
```diff
@@ -0,0 +1,64 @@
...
+      hosts:
+        - host: spire-tornjak-backend.$APP_SUBDOMAIN
+          paths:
+            - path: /
+              pathType: Prefix
+
+  ingress:
+    enabled: true
+    annotations:
+      route.openshift.io/termination: "passthrough"
```
kfox1111, line 24: In the other repo, we have some examples that are per ingress controller type..... maybe we should name things that way here since this looks to be specific to openshift?

examples/tornjak/README.md
```diff
@@ -40,3 +40,13 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
 You can now access Tornjak at [localhost:3000](http://localhost:3000).
 
 See [values.yaml](./values.yaml) for more details on the chart configurations to achieve this setup.
+
+## Tornjak and Ingress
+
+Obtain the Ingress subdomain (this process is cloud provider specific) and assign it
+to `APP_SUBDOMAIN` env. variable, then run the deployment with your values files:
```
marcofranssen, line 47:
```suggestion
to the `APP_SUBDOMAIN` env. variable, then run the deployment with your values files:
```

examples/openshift/openshift-values.yaml
```diff
@@ -5,12 +5,17 @@ global:
       enabled: true
```
kfox1111, line 5: Seems like all the changes in openshift-values.yaml are no longer needed? They were removed in the ingress controller type openshift pr.

examples/tornjak/README.md
```diff
@@ -40,3 +40,39 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
...
 See [values.yaml](./values.yaml) for more details on the chart configurations to achieve this setup.
+
+## Tornjak and Ingress
+
+Obtain the Ingress subdomain (this process is cloud provider specific) and assign it
+to the `APP_SUBDOMAIN` env. variable, then run the deployment with your values files:
+
+```shell
+export APP_SUBDOMAIN=$(kubectl get dns cluster -o jsonpath='{ .spec.baseDomain }')
+```
```
kfox1111, line 51: This I think only works on certain clusters. It doesn't work for crc / minikube I think.

Also, it should probably print it out?

examples/tornjak/README.md
```diff
@@ -40,3 +40,39 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
...
+Then run the deployment with your values files:
+
+```shell
+helm upgrade --install --namespace spire-server spire charts/spire \
+--values examples/production/values.yaml \
+--values examples/tornjak/values.yaml \
+--set spiffe-csi-driver.kubeletPath=/var/data/kubelet \
+--set spiffe-csi-driver.restrictedScc.enabled=true \
+--values examples/production/example-your-values.yaml \
+--values examples/tornjak/values-ingress.yaml \
```
kfox1111, line 66: Would really rather the user not update tornjak/values-ingress.yaml with custom things, when they could/should just do example-your-values.yaml. Maybe reorder so that example-your-values.yaml is last and suggest that file be edited instead?

examples/tornjak/values-ingress.yaml
```diff
@@ -0,0 +1,41 @@
+global:
+  spire:
+    jwtIssuer: https://oidc-discovery.$APP_SUBDOMAIN
```
kfox1111, line 3: With all the rest of the ingress pr's in place and examples-your-values.yaml, I think the only thing left in this file that is needed in this file is the bits that enable various ingress's?

examples/tornjak/README.md
```diff
@@ -40,3 +40,39 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
...
+_Note: The location of the apps subdomain may be different in certain environments_
+
+Then run the deployment with your values files:
+
+```shell
+helm upgrade --install --namespace spire-server spire charts/spire \
+--values examples/production/values.yaml \
+--values examples/tornjak/values.yaml \
+--set spiffe-csi-driver.kubeletPath=/var/data/kubelet \
+--set spiffe-csi-driver.restrictedScc.enabled=true \
```
kfox1111, line 64: I think these options are for ibm openshift only? Maybe drop it for this example?

examples/openshift/openshift-values.yaml
```diff
@@ -46,6 +45,10 @@ spiffe-oidc-discovery-provider:
     runAsUser: null
     runAsGroup: null
     fsGroup: null
+  securityContext:
+    allowPrivilegeEscalation: true
+    runAsNonRoot: false
+    privileged: true
```
kfox1111, line 51: Why did you need these bits? I was able to do it on openshift local without it. How does it break?

mrsabath: You are correct. I just test it and removed it.

kfox1111:
Had to force push to fix the DCO issue github created.

examples/tornjak/README.md
```diff
@@ -40,3 +40,46 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
 You can now access Tornjak at [localhost:3000](http://localhost:3000).
 
 See [values.yaml](./values.yaml) for more details on the chart configurations to achieve this setup.
+
+## Tornjak and Ingress
```
kfox1111, line 44: Since the openshift varient is documented below, perhaps we make this bit work for vanilla clusters?
```suggestion
## Tornjak and Ingress with ingress-nginx
```

examples/tornjak/README.md
```diff
@@ -40,3 +40,46 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
...
+
+Obtain the Ingress subdomain (this process is cloud provider specific) and assign it
+to the `APP_SUBDOMAIN` env. variable, then run the deployment with your values files:
+
+```shell
+export APP_SUBDOMAIN=$(kubectl get dns cluster -o jsonpath='{ .spec.baseDomain }')
+echo $APP_SUBDOMAIN
+```
```
kfox1111, line 52:
```suggestion
```

examples/tornjak/README.md
```diff
@@ -40,3 +40,46 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
...
+
+Update the `values-ingress.yaml` file with your subdomain.
+
+_Note: The location of the apps subdomain may be different in certain environments_
+
+Then run the deployment with your values files:
```
kfox1111, line 58:
```suggestion
Update examples/production/example-your-values.yaml with your information, most importantly, trustDomain.
```

examples/tornjak/README.md
```diff
@@ -40,3 +40,46 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
...
+
+```shell
+helm upgrade --install --namespace spire-server spire charts/spire \
+--values examples/production/values.yaml \
+--values examples/tornjak/values.yaml \
+--values examples/production/example-your-values.yaml \
+--values examples/tornjak/values-ingress.yaml \
```
kfox1111, line 65:
```suggestion
--values examples/tornjak/values-ingress.yaml \
--set global.spire.ingressControllerType=ingress-nginx \
--values examples/production/example-your-values.yaml \
```

kfox1111 (approved):
LGTM! Thanks for working through this! :)

examples/tornjak/README.md
```diff
@@ -41,3 +41,35 @@ kubectl -n spire-server port-forward service/spire-tornjak-frontend 3000:3000
...
+--render-subchart-notes --debug
+```
+
+## Tornjak and Ingress on Openshift
+
+When deploying on Openshift, follow the deployment setup as described in
+[Openshift README](../openshift/README.md)
+
+Then just add Openshift specific configuration to the above command:
+```shell
```
marcofranssen, line 65:
```suggestion

```shell
```

PR #34, opened 2023-10-12
Prune old branch before creating new one

faisal-memon:
Need to prune the local cached copy of bump-spire-version otherwise get an error when running the release script:

```
 $ ./release-chart.sh --chart spire --current-version 0.13.0 --new-version 0.13.1 --dry-run
Already on 'main'
Your branch is up to date with 'origin/main'.
Already up to date.
branch 'bump-spire-version' set up to track 'main'.
Switched to a new branch 'bump-spire-version'
[bump-spire-version b2cb089] Bump spire Helm Chart version from 0.13.0 to 0.13.1
 2 files changed, 2 insertions(+), 2 deletions(-)
To https://github.com/spiffe/helm-charts-hardened.git
 ! [rejected]        bump-spire-version -> bump-spire-version (stale info)
error: failed to push some refs to 'https://github.com/spiffe/helm-charts-hardened.git'
aborted: you must first push the current branch to a remote, or use the --head flag
```


marcofranssen:
This isn't required if we let GH auto remove merged branches. We might have forgotten to do that toggle when we switched repo.

Probably better to not do this automatically to prevent overwriting an open release.

faisal-memon:
@marcofranssen This is for the local cached copy I believe. GH is set to auto remove merged branches.

marcofranssen:
For local we do the checkout with -B instead of -b that already does this.

-r is for remote branches as well origin points to remote

marcofranssen:
I see you are working via a fork. That might be the issue.

faisal-memon:
@marcofranssen for the releases i don't work off fork.

marcofranssen:
This isn't required.

```shell
git push -u origin --force-with-lease
```

This line on line 116 in the script already pushes the branch corrctly. There is no need to remove it upfront.

The line

```
git checkout -B ...
```

Ensures the branch locally is forcefully recreated.

What are you trying to resolve with this? What is the error you are seeing? All should already work as expected.

marcofranssen:
@faisal-memon do you agree closing this one with my explanation from previous comment?

PR #39, opened 2023-10-17
Fix Tornjak HTTPS ingress port

mrsabath:


charts/spire/charts/spire-server/templates/tornjak-ingress.yaml
```diff
@@ -12,5 +12,9 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  {{ include "spire-lib.ingress-spec" (dict "ingress" .Values.tornjak.ingress "svcName" (include "spire-tornjak.servicename" .) "port" .Values.tornjak.service.ports.http) | nindent 2 }}
+  {{- if .Values.tornjak.ingress.tlsBackendPort -}}
```
kfox1111, line 15: Somehow this flag got merged without it being used? I don't see any reference to it outside of values/readme in tree.

Thinking the user may set up mtls but forget to set this flag and then are exposed to a security incident?

Maybe it should use this output instead?
```
(include "spire-tornjak.connectionType" .)
```

Then we can get rid of tlsBackendPort all together and have one less thing for the user to configure.

kfox1111 (approved):
Looks good to me. Thanks! :)

faisal-memon (approved):

PR #40, opened 2023-10-17
Update to SPIRE 1.8.2

### LINKED ISSUE #12: Update to SPIRE 1.8.0

### LINKED ISSUE #25: Drop spire version ranges
Here: https://github.com/spiffe/helm-charts-hardened#dependencies-and-version-compatibility

we claim to support all the 1.6.x and 1.7.x series. We don't test hardly any of it though. Thinking we should only claim official support of the version we are testing/releasing a particular version of the chart with?

faisal-memon:
Updates to SPIRE 1.8.2. Looking throught https://github.com/spiffe/spire/issues/2670 it mentions backwards compatibility. If no profile is specified it imples `profile "https_spiffe" {}`.

fixes #12 
fixes #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Unless otherwise noted in an application chart README, the following dependencie
 
 | Dependency | Supported Versions |
 |:-----------|:-------------------|
-| SPIRE      | `1.6.x`, `1.7.x`   |
+| SPIRE      | `1.8.2`            |
 | Helm       | `3.x`              |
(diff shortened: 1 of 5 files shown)


kfox1111:
Ok, so, reading through the both pr's and trunk, it looks like we're only supporting profile "https_spiffe" (implicitly) today, so we won't be breaking things if we move forward. But we do need to support the other options at some point.

kfox1111 (approved):
LGTM

PR #13, opened 2023-10-03
Support Openshift deployment 

mrsabath:
This PR brings the changes required to support SPIRE charts deployment on OpenShift version 4.13 or higher 

fixes #69

PR #51, opened 2023-10-20
Fix Tornjak HTTP/HTTPS port values

mrsabath:
The port values were incorrectly defined


mrsabath:
This will be fixed by the Openshift PR #13

PR #48, opened 2023-10-19
Simple ingress support

kfox1111:
For a lot of situations, ingress rules can be derived from trustDomain. Change the code to generate rules by default for the user.

PR #52, opened 2023-10-21
Ingress type openshift

kfox1111:


kfox1111:
Resolved conflicts. Removed redundant annotation because openshift will do it by default.

kfox1111:
Tested and works on top of openshift local.

Depends on https://github.com/spiffe/helm-charts-hardened/pull/48 merging first.

mrsabath (approved):
I tested this PR on Openshift cluster running in IBM Cloud. I setup the ingress (without TLS), enabled Tornjak, and OIDC. Then started a demo container and everything works as expected. All the services are accessible via ingress. 
LGTM!

PR #59, opened 2023-10-25
Added environment variable to spire-agent to inject node name

sabre1041:
Introduce injecting node name into spire agent instead of calling kubelet on localhost


kfox1111:
What does this change? MY_NODE_NAME is just an environment variable. Whats using it?

kfox1111:
oh. here we go:
https://github.com/spiffe/spire/blob/main/pkg/agent/plugin/workloadattestor/k8s/k8s.go#L42

does openshift not set the hostname on the pod to the hosts hostname like vanilla k8s on a net=host pod?

charts/spire/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: spire
 description: >
   A Helm chart for deploying the complete Spire stack including: spire-server, spire-agent, spiffe-csi-driver, spiffe-oidc-discovery-provider and spire-controller-manager.
 type: application
-version: 0.14.0
+version: 0.15.0
```
faisal-memon, line 6: Version bump is not needed.

charts/spire/README.md
```diff
@@ -1,6 +1,6 @@
 # spire
 
-![Version: 0.14.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.7.2](https://img.shields.io/badge/AppVersion-1.7.2-informational?style=flat-square)
+![Version: 0.15.0](https://img.shields.io/badge/Version-0.15.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.7.2](https://img.shields.io/badge/AppVersion-1.7.2-informational?style=flat-square)
```
faisal-memon, line 3: Same as above.

charts/spire/Chart.yaml
mrsabath, line 6:
```suggestion
version: 0.14.0
```

kfox1111:
oh... its not hostname related, but how the agent talks to kubelet... default if unset is:
https://github.com/spiffe/spire/blob/main/pkg/agent/plugin/workloadattestor/k8s/k8s.go#L519-L522 127.0.0.1

so, depending on how the hosts firewall is setup, changing it to hostname could be a breaking change....

we should probably make it setable but still default to to the existing behavior for now...

charts/spire/README.md
mrsabath, line 3:
```suggestion
![Version: 0.14.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.7.2](https://img.shields.io/badge/AppVersion-1.7.2-informational?style=flat-square)
```

mrsabath (commented):
Bringing back the versions

faisal-memon:
> oh... its not hostname related, but how the agent talks to kubelet... default if unset is: https://github.com/spiffe/spire/blob/main/pkg/agent/plugin/workloadattestor/k8s/k8s.go#L519-L522 127.0.0.1
> 
> so, depending on how the hosts firewall is setup, changing it to hostname could be a breaking change....
> 
> we should probably make it setable but still default to to the existing behavior for now...

Ive run into this issue before where the spire agent was not able to make the DNS request because of firewall issues. I agree with making it setable.

kfox1111:
So, that leaves discussion around it being a potential breaking change left....

Does openshift not allow 127.0.0.1 to kubelet out of the box? Trying to figure out how to get a good default, or at least some logic in here for it without breaking existing folks.

faisal-memon:
> So, that leaves discussion around it being a potential breaking change left....
> 
> Does openshift not allow 127.0.0.1 to kubelet out of the box? Trying to figure out how to get a good default, or at least some logic in here for it without breaking existing folks.

I think a simple bool like `nodeNameFromSpec` default to false should do it. The new section can be gated on the bool being set.

kfox1111:
> Easiest would be to enable a new "env" value where the end user can define their own environment variables. It would be the default in the set of openshift values but would remain as prior logic elsewhere unless they would want to opt in

Yeah. but I'm thinking it may be useful outside of openshift....
maybe a value like spire-agent.kubelet.contactByHostname true/false, default it to false everywhere but forced true if global.openshift=true? there ever a reason to use 127.0.0.1 still with openshift?

sabre1041:
@kfox1111 The reason for the original approach is that it follows the same pattern used by the csi-driver chart

https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spiffe-csi-driver/templates/daemonset.yaml#L50

kfox1111:
> @kfox1111 The reason for the original approach is that it follows the same pattern used by the csi-driver chart
> 
> https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spiffe-csi-driver/templates/daemonset.yaml#L50

Hmm.... I think the usage there is slightly different though. its used as a unique name, but it isn't contacted over tcp, just unix sockets. so won't have the dns/firewall issues that tcp to kubelet would have with the proposed change.

marcofranssen:
I'll test this tomorrow on my test cluster, to see if there is a breaking change at all. I think this might just work with backwards compatibility.

In the end spec.nodeName should be the same as the value currently retrieved I guess.

kfox1111:
> I'll test this tomorrow on my test cluster, to see if there is a breaking change at all. I think this might just work with backwards compatibility.
> 
> In the end spec.nodeName should be the same as the value currently retrieved I guess.

Totally depends on the firewall on the host, so you cant test them all. I do have some hosts I think it will break on. I have some pretty tight firewalls.

kfox1111:
And there's this bit of logic:
https://github.com/spiffe/spire/blob/main/pkg/agent/plugin/workloadattestor/k8s/k8s.go#L471-L496
that is otherwise inaccessible I think.

no, I think it has to be optional rather then always just set.

sabre1041:
> And there's this bit of logic: https://github.com/spiffe/spire/blob/main/pkg/agent/plugin/workloadattestor/k8s/k8s.go#L471-L496 that is otherwise inaccessible I think.
> 
> no, I think it has to be optional rather then always just set.

What approach should we use? the `env` or the `nodeNameFromSpec`? Personally, im favoring the `env` approach

kfox1111:
when you say env, do you mean, adding an extraEnv setting to https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-agent/values.yaml and then setting it via values?

sabre1041:
> when you say env, do you mean, adding an extraEnv setting to https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-agent/values.yaml and then setting it via values?

Yup exactly

kfox1111:
IMO, we should have extraEnv for that chart anyway, but don't currently it looks like. So, one way or another we probably should add that part.

faisal-memon:
extraEnv works for me.

kfox1111:
K. lets go with that  for now, and can revisit usability improvements later?

charts/spire/Chart.yaml
kfox1111, line 6:
```suggestion
version: 0.14.0
```

charts/spire/README.md
kfox1111, line 3:
```suggestion
![Version: 0.14.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.7.2](https://img.shields.io/badge/AppVersion-1.7.2-informational?style=flat-square)
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -71,6 +71,10 @@ spec:
           image: {{ template "spire-lib.image" (dict "appVersion" $.Chart.AppVersion "image" .Values.image "global" .Values.global) }}
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args: ["-config", "/run/spire/config/agent.conf"]
+          {{- if gt (len .Values.extraEnvVars) 0 }}
```
kfox1111, line 74:
```suggestion
          {{- with .Values.extraEnvVars }}
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -71,6 +71,10 @@ spec:
...
+          env:
+          {{- toYaml .Values.extraEnvVars | nindent 12 }}
```
kfox1111, line 76:
```suggestion
          {{- toYaml . | nindent 12 }}
```

kfox1111 (commented):
Some comments to fix the version, and some tweaks to make it looks similar to the rest of the chart in suggestions. Then, it looks good to me. Thanks! :)

charts/spire/Chart.yaml
mrsabath: Bumps are automated as a separate process

kfox1111: We batch up a bunch of changes in main, then bump the version during release.

charts/spire/charts/spire-agent/templates/daemonset.yaml
kfox1111: yeah, consistency within the charts has not been kept very well. Trying to clean it up as we go.

kfox1111: I'm ok skipping this part for now too though.

sabre1041:
@kfox1111 Should be updated to not change version of charts. Also did not change how the `extraEnvVars` are being injected as charts will uniformly be refactored later on

charts/spire/README.md
```diff
@@ -1,6 +1,6 @@
 # spire
 
-![Version: 0.14.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.7.2](https://img.shields.io/badge/AppVersion-1.7.2-informational?style=flat-square)
+![Version: 0.14.0](https://img.shields.io/badge/Version-0.14.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.7.2](https://img.shields.io/badge/AppVersion-1.7.2-informational?style=flat-square)
```
kfox1111, line 3:
```suggestion
![Version: 0.14.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.7.2](https://img.shields.io/badge/AppVersion-1.7.2-informational?style=flat-square)
```

kfox1111 (approved):
LGTM

marcofranssen (approved):
LGTM 🚀

PR #60, opened 2023-10-25
spire-controller-manager 0.4.0 support

kfox1111:
Updates image and CRD's to 0.4.0. Adds autoPopulateDNSNames and className support which were added in 0.4.0.


kfox1111:
Depends on this already merged feature getting a release: https://github.com/spiffe/spire-controller-manager/pull/230

charts/spire/charts/spire-server/values.yaml
```diff
@@ -370,6 +370,14 @@ notifier:
 controllerManager:
   ## @param controllerManager.enabled Flag to enable controller manager
   enabled: false
+
+  ## @param controllerManager.autoClassName if true, the controller class name will be Release.Namespace-Release.Name to automatically not conflict with other installs, allowing parallel installs.
+  autoClassName: true
+  ## @param controllerManager.className specify to use an explicit class name
+  className: ""
+  ## @param controllerManager.watchClassless specify to process custom resources without class name specified. Useful to migrate to class names from classless installs
```
faisal-memon, line 378: Can we follow the same pattern as the ingress, that we auto fill in the className if its empty string?

kfox1111: Done

charts/spire/README.md
```diff
@@ -32,6 +32,16 @@ For production installs, please see [the production example](https://github.com/
 
 ## Upgrade notes
 
+### 0.15.X
+
+The spire-crds chart has been updated. Please ensure you have upgraded spire-crds before upgrading the spire chart.
+
+The chart now includes support for multiple parallel installs of the spire-controller-manager. To facilitate this change, all custom resources now take
+a className property. Out of the box the chart now uses and requires it to be set. If you have not loaded any yourself, the upgrade process will be transparent.
```
marcofranssen, line 40: > If you have not loaded any yourself

What is `any` here?

> uses and requires it to be set

Assuming the className has to be set? Can't we make it such it has a default for the simple spire-system way of installing all in one namespace, so you only have to override for the advanced split namespace install?

> className

What does a className mean? A class of what? Looking at this as a complete newby to the feature so miss the context of the entire feature, its working and the lingo used.

charts/spire/README.md
```diff
@@ -32,6 +32,16 @@ For production installs, please see [the production example](https://github.com/
...
+
+If you have loaded your own custom resources, set spire-server.controllerManager.watchClassless=true until you can update your custom resources to have the
+className for the instance specified.
```
marcofranssen, line 43: I think we need to explain this new feature in the first paragraph so people can follow allong. I have difficulty following this entire section as I miss the context of the entire feature.

charts/spire/charts/spire-server/README.md
```diff
@@ -233,6 +235,7 @@ In order to run Tornjak with simple HTTP Connection only, make sure you don't cr
...
 | `controllerManager.identities.federatesWith`                     | Other Spire server URLs for identity federation                                                                                                                                                                                                                                                                                                                      | `[]`                                                                                           |
+| `controllerManager.identities.downstream`                        | set if this is a root server                                                                                                                                                                                                                                                                                                                                         | `false`                                                                                        |
```
marcofranssen, line 238: Shouldn't the default be true, so it works out of the box for the simple usecase where you just have a single server?

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -217,3 +217,11 @@ The code below determines what connection type should be used.
 {{-   end }}
 {{ $args | toYaml }}
 {{- end -}}
+
+{{- define "spire-server.controller-manager-class-name" -}}
```
marcofranssen, line 221: Looking at this wouldn't instanceName be a better name?

Assuming this represents an instance of a spire-server.

charts/spire/charts/spire-server/templates/controller-manager-cluster-ids.yaml
```diff
@@ -24,5 +24,9 @@ spec:
   dnsNameTemplates:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  className: {{ include "spire-server.controller-manager-class-name" $root | quote}}
+  {{- if .identities.downstream }}
+  downstream: true
+  {{- end }}
```
marcofranssen, line 30:
```suggestion
  downstream: {{ .identities.downstream }}
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -370,6 +370,12 @@ notifier:
 controllerManager:
   ## @param controllerManager.enabled Flag to enable controller manager
   enabled: false
+
+  ## @param controllerManager.className specify to use an explicit class name. If empty, it will be automatically set to Release.Namespace-Release.Name to not conflict with other installs, enabling parallel installs.
+  className: ""
```
marcofranssen, line 375: instanceName would be a better name IMHO. className has a completely different meaning to me, while so far I see this is used as the name of a spire deployment instance.

marcofranssen (changes requested):
Couple of suggestions on renaming some stuff and adding more context for understanding the feature. I think people new to the feature (like me) will be unable to follow allong.

charts/spire/README.md
marcofranssen: Having reached the end of the entire PR I think instanceName would be much better naming as it is about the instance of a spire-server installation.

charts/spire/README.md
kfox1111: An attempt at that was:
> The chart now includes support for multiple parallel installs of the spire-controller-manager.

Should we put more description then that int he upgrade notes, or a separate section? Was trying to keep the upgrade notes concise for those not interested in the feature.

charts/spire/charts/spire-server/README.md
kfox1111: Downstream means that the cluster spire identity will be a downstream of this instance and therefore given intermediate CA's. So, only should be set on root spire instances.

charts/spire/charts/spire-server/templates/_helpers.tpl
kfox1111: Following the established k8s pattern. className is very common and I don't think we should deviate from it.

charts/spire/charts/spire-server/values.yaml
kfox1111: too much history I think to change, and instanceName is not clear either, as your putting it on an instance of a CR. So, your naming the CR a second time?

.github/tests/dependencies/spire-root-server-values.yaml
```diff
@@ -6,7 +6,14 @@ global:
 
 spire-server:
   controllerManager:
-    enabled: false
+    identities:
+      namespaceSelector:
+        kubernetes.io/metadata.name: spire-server
```
marcofranssen, line 11: in the simple setup without namespace splitting we tend to use `spire-system`. What is the best default here?

charts/spire/charts/spire-server/README.md
marcofranssen: That is my point. If I just have a simple setup, I should not require to set any value.

I think it is better to set it only when running downstream servers to keep the simple setup with proper defaults.

charts/spire/charts/spire-server/templates/_helpers.tpl
marcofranssen: Do you have a reference to that common pattern of className usage? I really have trouble understanding this naming. Hope I can read up somewhere a bit to understand this naming convention.

charts/spire/charts/spire-server/values.yaml
marcofranssen: I guess I completely lack the context on these changes, I am completely lost.

charts/spire/charts/spire-server/README.md
kfox1111: thats my point. in a default setup, it is correct being false. you only need to set it as true, if your setting up a nested deployment and the release is the root one.

charts/spire/charts/spire-server/templates/_helpers.tpl
kfox1111: Example of storage class useage:
https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/#using-dynamic-provisioning

Example of ingress class usage:
https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

StorageClass details:
https://kubernetes.io/docs/concepts/storage/storage-classes/

Ingress api documentation around ingress class:
https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class

Ingress-nginx charts configurables for classes:
https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml#L76-L81
https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml#L108-L125

.github/tests/dependencies/spire-root-server-values.yaml
kfox1111: This is configuring the root spire server to provide an identity for the leaf spire-server-0 pod in the spire-server namespace.

Its a clean way to take the place of the hacky thing we had before:
https://github.com/spiffe/helm-charts-hardened/blob/main/examples/nested/run-tests.sh#L36-L38

charts/spire/charts/spire-server/values.yaml
```diff
@@ -395,7 +401,7 @@ controllerManager:
     repository: spiffe/spire-controller-manager
     pullPolicy: IfNotPresent
     version: ""
-    tag: "0.3.0"
+    tag: "nightly"
```
kfox1111, line 404:
```suggestion
    tag: "0.4.0"
```

kfox1111:
@marcofranssen What would be the best path forward here? Would it help to have a call?

charts/spire/charts/spire-server/values.yaml
faisal-memon: @marcofranssen The context is this pr in controller manager: https://github.com/spiffe/spire-controller-manager/pull/230

charts/spire/README.md
```diff
@@ -32,6 +32,17 @@ For production installs, please see [the production example](https://github.com/
...
+### 0.15.X
+
+The spire-crds chart has been updated. Please ensure you have upgraded spire-crds before upgrading the spire chart.
+
+The chart now includes support for multiple parallel installs of the spire-controller-manager. To facilitate this change, all spire custom resources now take
+a className property. Out of the box the chart now uses and requires it to be set. If you have not loaded any spire custom resources yourself, the upgrade
+process will be transparent.
+
+If you have loaded your own spire custom resources, set spire-server.controllerManager.watchClassless=true until you can update your spire custom resources to
+have the className for the instance specified.
```
faisal-memon, line 44:
```suggestion
The chart now supports multiple parallel installs of spire-controller-manager. Each install will handle all custom resources with a matching `className` field.  By default this is set to `Release.Namespace-Release.Name` and the controller manager will only pick up custom resources with this className. 

If you have not loaded any SPIRE custom resources yourself, the upgrade process will be transparent. If you have loaded your own SPIRE custom resources, set `spire-server.controllerManager.watchClassless=true` until you can update your SPIRE custom resources to have the `className` for the instance specified.
```

examples/nested/run-tests.sh
```diff
@@ -33,10 +33,6 @@ helm upgrade --install --create-namespace spire charts/spire \
   --values "${DEPS}/spire-root-server-values.yaml" \
   --wait
 
-kubectl get nodes -o go-template='{{range .items}}{{printf "%s\n" .metadata.uid}}{{end}}' | while read -r line; do
-  kubectl exec -t spire-server-0 -n "spire-root-server" -- spire-server entry create -spiffeID spiffe://production.other/production/nested-spire -parentID "spiffe://production.other/spire/agent/k8s_psat/production/$line" -selector k8s:pod-label:app.kubernetes.io/name:server -downstream
-done
-
```
faisal-memon, line 39: Why is this being removed?

charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: {{ include "spire-controller-manager.fullname" . }}-webhook
+  name: {{ .Release.Namespace }}-{{ include "spire-controller-manager.fullname" . }}-webhook
```
faisal-memon, line 5: Are you planning to deploy the parallel controller managers in different namespaces?

faisal-memon:
@kfox1111 Overall looks good. Thanks for adding the other missing controller manager features. Just some minor questions inline.

charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
kfox1111: Yes.

For one cluster, I'm currently planning on 3.
spire-root - root ca for all other spires.
spire-server - regular examples/production deployment for the k8s cluster its on, but ca is from spire-root
spire-external - exported out of the cluster for non k8s workloads. ca is from spire-root.

Each instance would be hardened different ways for their particular use cases.

examples/nested/run-tests.sh
kfox1111: With the ability to have multiple controller managers, the spire-root server now has its own, and can be configured via 
the spire charts built in ClusterSPIFFEID rather then having to manually load spire entries into the database. Its now configured here:
https://github.com/spiffe/helm-charts-hardened/blob/controller-manager-class/.github/tests/dependencies/spire-root-server-values.yaml#L8-L16

This is much better then the for loop, as the loop is static. It only loads the known k8s nodes available at the time it runs, where as the controller manager will pick up new nodes and remove old nodes as they come and go from the k8s cluster.

charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
faisal-memon: the release name would be the same for all 3?

kfox1111: Yeah, that's how we typically name things. For singletons in their own namespaces, release name = chart name. The namespace differentiates things.

We've got other places in the main branch where we have added the .Release.Namespace into global objects to prevent naming conflict. This one was just missed as we didn't support multiple controller-managers til now.

examples/nested/run-tests.sh
marcofranssen: the downstream setting there is confusing to me? what has a root server to do with a downstream? Shouldn't it be the case nested servers should only refer to upstream servers. To not end up in chicken and egg scenarios?

E.g.

I start with a single spire instance, now later on I want to add a nested one, why would I have to change the upstream deployment in order to get the new nested (downstream) working?

charts/spire/charts/spire-server/templates/controller-manager-webhook.yaml
marcofranssen: For RBAC reasons I get the splitting of namespaces? But what if someone decides to put all in the same namespace with different release names? Should we support that?

kfox1111: That should work as well I think? The full name of the release is included in the name, so should be unique when multiple are in the same namespace.

examples/nested/run-tests.sh
kfox1111: Say you have a nested setup. you have essentially:
root spire-server -> root clusterspiffeid config -> upstream spire-agent -> nested spire-server -> nested spire clusterspiffeid config -> downstream spire-agent -> user workload.

nested spire-server is a downstream server of root spire-server. Downstream=true is set on root spire clusterspiffeid so that root spire-server will hand out an intermediate ca identity to the nested spire-server via upstream spire-agent.

nested clusterspiffeid does not have downstream true, so user workload won't get an intermediate ca.

marcofranssen: Ok it starts to become more clear for me. Guess this is good enough for now. In case we face issues or more questions from community we can always revise to make it more clear. Maybe it is just me 😊

marcofranssen (approved):
LGTM 🚀 Lets get some community feedback on this feature and its UX from a Helm perspective.

charts/spire/charts/spire-server/templates/pre-upgrade-hook.yaml
```diff
@@ -1,4 +1,9 @@
 {{- if eq ((dig "installAndUpgradeHooks" "enabled" .Values.controllerManager.installAndUpgradeHook.enabled .Values.global) | toString) "true" }}
+{{- $webhookname := printf "%s-%s-webhook" .Release.Namespace (include "spire-controller-manager.fullname" .) }}
+{{- $oldwebhookname := printf "%s-webhook" (include "spire-controller-manager.fullname" .) }}
+{{- if not (lookup "admissionregistration.k8s.io/v1" "ValidatingWebhookConfiguration" "" $webhookname) }}
+{{- $webhookname = $oldwebhookname }}
+{{- end }}
```
drewwells, line 6: These short term object renames need to be major releases. It's impossible to upgrade from 0.14.0 to 0.20.0 due to this rename logic being removed somewhere b/t here and 0.20.0

PR #67, opened 2023-10-29
Fix incorrect nesting of priorityClassName

InverseIntegral:
In my recent PR contributed to the old repository (https://github.com/spiffe/helm-charts/pull/480) I accidentally introduced a bug where the `priorityClassName` would only be applied if there either was an init container defined or if cert-manager was being used. If you want I can also add a test case for this particular bug. I would also like to fix this bug in https://github.com/spiffe/helm-charts-flex but I'm not sure if they are already accepting contributions :slightly_smiling_face:


kfox1111 (approved):
Good catch. Fix looks good to me. Thanks!

faisal-memon (approved):

PR #70, opened 2023-10-31
fix(spire-server): tornjak ingress port

### LINKED ISSUE #69: tornjak-backend ingress error
When using `spire-server.tornjak.ingress.enabled = true` helm wants to create an ingress with

```
spec:
  rules:
    - host: <>
      http:
        paths:
          - path: <>
            pathType: <>
            backend:
              service:
                name: <>
                port:
                  number: tornjak-srv-http
```
See tornjak ingress template: https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-server/templates/tornjak-ingress.yaml#L26C1-L27C1
See ingress template in spire-lib: https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/templates/_spire-lib.tpl#L125

This results in nginx ingress controller errors, because the number has to be an integer and not a string.

simonostendorf:
Closes #69

### DIFF
--- a/charts/spire/charts/spire-server/templates/tornjak-ingress.yaml
+++ b/charts/spire/charts/spire-server/templates/tornjak-ingress.yaml
@@ -23,8 +23,8 @@ metadata:
   {{- end }}
 spec:
   {{- if eq (include "spire-tornjak.connectionType" .) "http" }}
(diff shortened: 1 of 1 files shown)


faisal-memon:
Thanks @simonostendorf. This will be fixed by #13, so will close this one.

PR #74, opened 2023-11-02
Patch Tornjak ingress

mrsabath:
There are 2 ingress annotations in the spire-server. We should be using the one specific to tornjak not the top-level for spire server


kfox1111 (approved):
Good catch. LGTM.

We may need some more testing checks to make sure breakage doesn't happen.

PR #77, opened 2023-11-04
Add missing ClusterSPIFFEID fields

kfox1111:
There are a few options in the CRD not available via the chart. Sync them to the chart.


charts/spire/charts/spire-server/templates/controller-manager-cluster-ids.yaml
```diff
@@ -24,5 +24,17 @@ spec:
   dnsNameTemplates:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- with .identities.workloadSelectorTemplates }}
+  workloadTSelectoremplates:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .identities.ttl }}
+  ttl: {{ . | quote }}
```
faisal-memon, line 32: Why the extra `with` statement instead of just `ttl: {{.identities.ttl | quote }}`?

kfox1111: If the ttl value is unset, it doesn't include the ttl field at all in the crd, and then spire will fall back to whatever the spire-server is set to use as a default ttl.

faisal-memon (approved):

Issue #76, opened 2023-11-03
1.0 possible breakage list


kfox1111:
We've tried hard to ensure backwards compatibility, a bit to the detriment of usability out of the box. Before 1.0.0 we may want to have one release where we change the defaults to be easier to use. This issue will track possible things to change.

kfox1111:
autoPopulateDNSNames maybe should be true rather then false.

kfox1111:
spire-server.upstreamAuthority.certManager.ca.create could be set to true

kfox1111:
Should jwt issuer default to include https:// ?

kfox1111:
consider switching default identifiers to include nested names so taking non nested clusters and scale it out to multiple containers doesn't break things.

kfox1111:
base hostpath socket name on release.namespace-release.name by default so they don't conflict.

kfox1111:
In spire-agent, the value
```
socketAlternate:
  image:
```

could be collapsed into a general tools container along with the permissions fix init container.

faisal-memon:
Fix up naming so there is consistency between `socket` and `sockets` for socket path configs. Only use plural if there are multiple sockets. Reference comment: https://github.com/spiffe/helm-charts-hardened/pull/234#discussion_r1483364838

kfox1111:
ca_subject section here:
https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-server/values.yaml#L279-L285

should be updated to follow naming convention.

kfox1111:
We should go through all the values and ensure they follow the convention. There are a few like `spire-server.k8sbundle`

kfox1111:
https://github.com/spiffe/helm-charts-hardened/issues/269

kfox1111:
List of vars that dont match convention:
```
spire-server.dataStore.plugin_data
spire-server.federation.refresh_hint
spire-server.ca_subject.common_name
spire-server.upstreamAuthority.certManager.issuer_name
spire-server.upstreamAuthority.certManager.issuer_group
spire-server.upstreamAuthority.certManager.kube_config_file
spire-server.notifier.k8sbundle
spire-server.nodeAttestor.k8sPsat

spire-agent.nodeAttestor.k8sPsat
spire-agent.sds.defaultSvidName
spire-agent.sds.disableSpiffeCertValidation
```

PR #78, opened 2023-11-04
Allow additional CRs to be managed by the chart

kfox1111:
Sometimes additional ClusterSPIFFEIDs and the other CRs are needed. Add support for the end user to manage those extra CRs via the chart.


charts/spire/charts/spire-server/templates/controller-manager-cluster-ids.yaml
```diff
@@ -5,7 +5,6 @@ apiVersion: spire.spiffe.io/v1alpha1
 kind: ClusterSPIFFEID
 metadata:
   name: {{ include "spire-controller-manager.fullname" $root }}-service-account-based
-  namespace: {{ include "spire-server.namespace" $root }}
```
kfox1111, line 8: Noticed during development that the namespace was specified, but the object is cluster wide so namespace is always ignored. Removed to be consistent with the newly added objects.

charts/spire/charts/spire-server/README.md
```diff
@@ -240,6 +240,9 @@ In order to run Tornjak with simple HTTP Connection only, make sure you don't cr
...
 | `controllerManager.identities.downstream`                        | Set if this spire instance is a root server and the workloads are downstream servers.                                                                                                                                                                                                                                                                                | `false`                                                                                        |
+| `controllerManager.extraCRs.clusterSPIFFEIDs`                    | Specify additional ClusterSPIFFEID objects.                                                                                                                                                                                                                                                                                                                          | `{}`                                                                                           |
```
faisal-memon, line 243: What do you think of making the existing `identities` an array  instead of creating a new extraCRs section for it?

kfox1111: Hmm...

If it was an array, the user couldn't update just one field due to the way helm overwrites arrays provided by the user. We could do it as a dictionary though.

So, something like:
```
identities:
  default:
    enabled: true
    spiffeIDTemplate: spiffe://{{ .TrustDomain }}/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}
    podSelector: {}
    namespaceSelector: {}
    dnsNameTemplates: []
    federatesWith: []
    workloadSelectorTemplates: []
    ttl: ""
    jwtTTL: ""
    admin: false
    downstream: false
  other:
    same-stuff: here
```
The code would be the same as extraCRs.clusterSPIFFEIDs just pointed at identities instead and we can get rid of the extra code to handle the default.

What would we want to do with the other CR's in that case? Should we make another nesting level under identities for it? :
```
identities:
  dynamic:
    default:
      ...
  static:
    foo:
      ...
  federated:
    foo:
      ...
```

Or hang that stuff somewhere outside of identities?

Another option would be to structure things the cleanest way that doesn't break backwards comparability and handle it in the future with https://github.com/spiffe/helm-charts-hardened/issues/76

kfox1111: Talked at the meeting. Use the crd name directly under identities. backwards compat breakage of moving things under default is ok.

kfox1111: Feedback incorporated. I think this is ready now.

charts/spire/README.md
```diff
@@ -32,6 +32,10 @@ For production installs, please see [the production example](https://github.com/
 
 ## Upgrade notes
 
+### 0.16.X
+
+The settings under "spire-server.controllerManager.identities" have all been moved under "spire-server.controllerManager.identities.clusterSPIFFEIDs.default". If you have changed any from the defaults, please update them to the new location during upgradey.
```
kfox1111, line 37:
```suggestion
The settings under "spire-server.controllerManager.identities" have all been moved under "spire-server.controllerManager.identities.clusterSPIFFEIDs.default". If you have changed any from the defaults, please update them to the new location during upgrade.
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -445,38 +445,66 @@ controllerManager:
     - local-path-storage
 
   identities:
```
faisal-memon, line 447: Do we still need this extra `identities:` layer?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -445,38 +445,66 @@ controllerManager:
...
+    ## @param controllerManager.identities.clusterFederatedTrustDomains Specify additional ClusterFederatedTrustDomain objects.
+    clusterFederatedTrustDomains: {}
+    # foo:
+    #   labels:
+    #     foo: bar
+    #   bundleEndpointProfile:
+    #     endpointSPIFFEID: spiffe://example.com/foo
+    #     type: https_spiffe
+    #   bundleEndpointURL: https://rootserver.example.com:1234
+    #   trustDomain: example.com
```
faisal-memon, line 507: The options won't get documented in the readme with everything commented out. Would it be better to have something but disabled for documentation purposes?

charts/spire/charts/spire-server/values.yaml
kfox1111: Oh, yeah. I think we could get rid of that.... Good idea. Will take a stab at that.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -445,38 +445,66 @@ controllerManager:
...
-    jwtTTL: ""
-    ## @param controllerManager.identities.admin Indicates any pod matched by this identity will be an admin. Use this with extreme care.
-    admin: false
-    ## @param controllerManager.identities.downstream Set if this spire instance is a root server and the workloads are downstream servers.
-    downstream: false
-    ## @param controllerManager.identities.autoPopulateDNSNames Auto populate DNS names from services attached to pods
-    autoPopulateDNSNames: false
+    clusterSPIFFEIDs:
+      default:
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.enabled Flag to enable default identities for controller manager
```
faisal-memon, line 450:
```suggestion
        ## @param controllerManager.identities.clusterSPIFFEIDs.default.enabled Enable this identity for controller manager
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -445,38 +445,66 @@ controllerManager:
...
+        ttl: ""
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.jwtTTL Indicates an upper-bound time-to-live for JWT SVIDs. If unset, the cluster default will be chosen.
+        jwtTTL: ""
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.admin Indicates any pod matched by this identity will be an admin. Use this with extreme care.
+        admin: false
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.downstream Set if this spire instance is a root server and the workloads are downstream servers.
+        downstream: false
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.autoPopulateDNSNames Auto populate DNS names from services attached to pods
+        autoPopulateDNSNames: false
+      # foo:
```
faisal-memon, line 482:
```suggestion
      # You can specific additional ClusterSPIFFEIDs following this example.
      # foo:
```

charts/spire/charts/spire-server/values.yaml
faisal-memon: nm, i think its good to have it as it keeps everything clustered together in the docs.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -445,38 +445,66 @@ controllerManager:
...
-    ## @param controllerManager.identities.autoPopulateDNSNames Auto populate DNS names from services attached to pods
-    autoPopulateDNSNames: false
+    clusterSPIFFEIDs:
+      default:
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.enabled Flag to enable default identities for controller manager
+        enabled: true
+
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.spiffeIDTemplate Spiffe ID template for identities
+        spiffeIDTemplate: spiffe://{{ .TrustDomain }}/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.podSelector [object] Selector for pods to issue identity
```
faisal-memon, line 455: Is there anyway to have the comments not have `default` in them; instead have something more generic like `[name]`.

PR #83, opened 2023-11-07
Fix the typo in the production ingress README

mrsabath:
I think we introduced the typo in the name convention for the production ingress files and the documentation.


faisal-memon:
@mrsabath I think the change is fine but the tests are failing

kfox1111:
Export / expose are basically synonyms. Expose has a negative connotation in security circles so I didn't chose that originally. But the more I think about your suggestion, the more I like it. It would make people think, "Do I really want to expose this service outside the cluster". Which is a good thought to have.

kfox1111 (approved):
LGTM! Thanks! :)

PR #84, opened 2023-11-08
Update to SPIRE 1.8.4

faisal-memon:
fixes: https://github.com/spiffe/helm-charts-hardened/issues/71

### DIFF
--- a/charts/spire/README.md.gotmpl
+++ b/charts/spire/README.md.gotmpl
@@ -19,7 +19,7 @@
 
 | Dependency | Supported Versions |
 |:-----------|:-------------------|
-| SPIRE      | `1.5.3+`, `1.6.3+` |
+| SPIRE      | `1.8.4`            |
 | Helm       | `3.x`              |
 | Kubernetes | `1.22+`            |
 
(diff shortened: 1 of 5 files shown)


charts/spire/README.md.gotmpl
```diff
@@ -19,7 +19,7 @@
 
 | Dependency | Supported Versions |
 |:-----------|:-------------------|
-| SPIRE      | `1.5.3+`, `1.6.3+` |
+| SPIRE      | `1.8.4`            |
```
kfox1111, line 22: I think this whole file may not be used anymore? Should we get rid of it?

kfox1111 (approved):
Looks good to me. One issue identified, but we could also do that in a followup PR.

charts/spire/README.md.gotmpl
marcofranssen: correct. This got replaced when we switched doc tool

marcofranssen (approved):
LGTM 🚀

charts/spire/README.md.gotmpl
kfox1111: Filed https://github.com/spiffe/helm-charts-hardened/issues/87

marcofranssen: https://github.com/spiffe/helm-charts-hardened/pull/86

PR #97, opened 2023-11-11
Documentation cleanup

kfox1111:
Fix links to the repo after move. Remove references to other versions of images we don't support.


faisal-memon (approved):
Thanks for doing this @kfox1111

PR #102, opened 2023-11-13
Add some nested diagrams

kfox1111:


examples/nested/README.md
```diff
@@ -0,0 +1,10 @@
+# Possible Nesting Configurations
+
+There are multiple ways of configuring the chart depending on what you want to use nesting for.
+
+## Diagram for nesting across Kubernetes clusters:
```
faisal-memon, line 5:
```suggestion
## Nesting across Kubernetes clusters:
```

examples/nested/README.md
```diff
@@ -0,0 +1,10 @@
...
+![Multiple Kubernetes Cluster](./multicluster.png)
+
+## Diagram using nesting within a Kubernetes cluster:
```
faisal-memon, line 8:
```suggestion
## Nesting within a Kubernetes cluster:
```

examples/nested/README.md
```diff
@@ -0,0 +1,10 @@
+# Possible Nesting Configurations
```
faisal-memon, line 1: Can you add a small blurb about using `namesapceOverride` to get the components into different namespaces?

A small pros/cons discussion of when to use the different approaches would be nice too.

kfox1111: Was trying to take an iterative approach here. Getting solid examples of these configurations will take quite a bit of time to get right and document. But the diagrams are useful to show off what is currently possible with the chart. So, maybe 2  pr's?

faisal-memon: Yea thats fine.

PR #103, opened 2023-11-14
Added ability to create namespaces

sabre1041:
Enables managing the creation of `spire-system` and `spire-server` namespaces


charts/spire/charts/spire-server/values.yaml
```diff
@@ -725,3 +725,24 @@ tests:
...
+## @section Namespaces
+namespaces:
+  ## @param namespaces.system.name Name of the Spire system Namespace.
+  ## @param namespaces.system.create Create a Namespace for Spire system resources.
+  ## @param namespaces.system.annotations [object] Annotations to apply to the Spire system Namespace.
+  ## @param namespaces.system.labels [object] Labels to apply to the Spire system Namespace.
+  system:
+    name: "spire-system"
+    create: false
+    annotations: {}
```
marcofranssen, line 738: I would set all the annotations from production example here by default. Especially because namespace creation is opt-in, that would only happen if the namespace creation is set to true.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -725,3 +725,24 @@ tests:
...
+    labels: {}
+  ## @param namespaces.server.name Name of the Spire server Namespace.
+  ## @param namespaces.server.create Create a Namespace for Spire server resources.
+  ## @param namespaces.server.annotations [object] Annotations to apply to the Spire server Namespace.
+  ## @param namespaces.server.labels [object] Labels to apply to the Spire server Namespace.
+  server:
+    name: "spire-server"
+    create: false
+    annotations: {}
```
marcofranssen, line 747: Same see above.

marcofranssen (changes requested):
This is great if this still works with the `--create-namespace` cli option. Haven't tested it myself just yet.

Things we need to test is the behavior for the following situation.

```shell
helm upgrade --install --create-namespace --namespace spire-server --set namespace.server.create=true spire spiffe/spire
```

as well the following scenario

```shell
kubectl create ns spire-server
helm upgrade --install --namespace spire-server --set namespace.server.create=true spire spiffe/spire
```

Based on the outcome of above we might need some additional documentation on the caveats when using this function. E.g. can the namespace exist before installing with these values, or can we use the `--create-namespace` flag.

I love the idea of this feature, which can drastically simplify installing the production example. Just would love to test the above usecases to see how it behaves with existing namespace or namespace created via the CLI option.

kfox1111:
@sabre1041 laid out a scenario where it would always work. Installing spire-crds and spire charts in a 3rd namespace, say, spire-releases, enabling both namespace creation with the chart, and using namespaceOverrides to land all the workloads into the appropriate namespaces. So, even if helm has a problem with installing the chart in spire-server and having the spire-server ns object in the chart, the pr is still 100% useful in some cases.

charts/spire/charts/spire-server/README.md
```diff
@@ -315,3 +315,16 @@ In order to run Tornjak with simple HTTP Connection only, make sure you don't cr
...
+| Name                            | Description                                         | Value          |
+| ------------------------------- | --------------------------------------------------- | -------------- |
+| `namespaces.system.name`        | Name of the Spire system Namespace.                 | `spire-system` |
+| `namespaces.system.create`      | Create a Namespace for Spire system resources.      | `false`        |
+| `namespaces.system.annotations` | Annotations to apply to the Spire system Namespace. | `{}`           |
+| `namespaces.system.labels`      | Labels to apply to the Spire system Namespace.      | `{}`           |
+| `namespaces.server.name`        | Name of the Spire server Namespace.                 | `spire-server` |
+| `namespaces.server.create`      | Create a Namespace for Spire server resources.      | `false`        |
+| `namespaces.server.annotations` | Annotations to apply to the Spire server Namespace. | `{}`           |
+| `namespaces.server.labels`      | Labels to apply to the Spire server Namespace.      | `{}`           |
```
kfox1111, line 330: Wondering if we should lift the values/templates up to charts/spire/values.yaml and charts/spire/templates/

They aren't really specific to the spire-server chart, and nest the settings deeply when it could just be:
namespaces.system.name right at the root of the values.

kfox1111: After further discussion, it probably should be a global under global.spire.namespaces.*

We can use that value space in other places to enable simpler deployment in follow on pr's.

kfox1111 (approved):
LGTM

charts/spire/charts/spire-server/values.yaml
kfox1111: Unfortunately, the annotations aren't always consistent between configurations. They are a bit different in openshift and vanilla k8s and even between openshift deployments. So, I think we will need to make it programatic.

I do like the idea of making it easy to do the best practice annotations though...
I have another pr thats rolling in the best practices from the examples into the chart. maybe we add another one for the namespace annotations, like:
https://github.com/spiffe/helm-charts-hardened/pull/106/files#diff-e5d95a142cf1134e77e5c83dfeac07f49d1f433253a8b07d11a652c784403a6bR7-R9
?

kfox1111:
Doing some testing on it. It does indeed have a problem if the namespace that its installed into is attempted to be created:
> $ helm install -n spire-server spire charts/spire --set global.spire.namespaces.system.create=true,global.spire.namespaces.server.create=true -f examples/production/values.yaml -f examples/production/example-your-values.yaml
> Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: Namespace "spire-server" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "spire"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "spire-server"

Tried a few variants of it.

marcofranssen (approved):
Approving as I see my concerns where answered. Will not merge as I see still some comments from @kfox1111 which haven't been answered, despite he already approved it, so leave the "Squash Merge" to @kfox1111 .

Issue #99, opened 2023-11-12
Upgrade Job always changes


DingGGu:
When helm upgrade, Job and pre-upgrade ClusterRole always seeing diff in 0.15.1

```patch
spire-system, spire-server-pre-upgrade, ClusterRole (rbac.authorization.k8s.io) has changed:
...
  apiVersion: rbac.authorization.k8s.io/v1
  kind: ClusterRole
  metadata:
    name: spire-server-pre-upgrade
    annotations:
      "helm.sh/hook": pre-upgrade
      "helm.sh/hook-delete-policy": before-hook-creation, hook-succeeded, hook-failed
  rules:
    - apiGroups: ["admissionregistration.k8s.io"]
      resources: ["validatingwebhookconfigurations"]
-     resourceNames: ["spire-system-spire-controller-manager-webhook"]
+     resourceNames: ["spire-controller-manager-webhook"]
      verbs: ["get", "patch"]
spire-system, spire-server-pre-upgrade, Job (batch) has changed:
...
        securityContext:
          {}
        containers:
        - name: post-install-job
          securityContext:
            {}
          image: docker.io/rancher/kubectl:v1.26.10
          args:
            - patch
            - validatingwebhookconfiguration
-           - spire-system-spire-controller-manager-webhook
+           - spire-controller-manager-webhook
```

faisal-memon:
Hi @DingGGu this is expected for this release. We added support for running multiple controller managers with https://github.com/spiffe/helm-charts-hardened/pull/60. They are differentiated by running in different namespaces so we changed the name of some of the webhooks to reflect that.

kfox1111:
I've done some testing locally. It seems like 'helm diff' may have a bug in it where it does not honor the 'lookup' function call that is done in the job to successfully upgrade from the previous version. When the helm upgrade happens, the correct name is used, so its kind of a false positive. We will get rid of the lookup call in the 0.16.0 release so the diff should be fixed then.

DingGGu:
Thank you! I didn't know the context of the past.

kfox1111:
Going to close this for now. If you need anything else, please let us know.

PR #104, opened 2023-11-15
Remove unneeded lookup function from upgrade hook

kfox1111:
Those upgrading to 0.16.0 should no longer need the code to check for the old webhook.
    
Fixes: https://github.com/spiffe/helm-charts-hardened/issues/99


charts/spire/charts/spire-server/templates/pre-upgrade-hook.yaml
```diff
@@ -73,7 +68,7 @@ spec:
         args:
           - patch
           - validatingwebhookconfiguration
-          - {{ $webhookname }}
+          - {{ .Release.Namespace }}-{{ include "spire-controller-manager.fullname" . }}-webhook
```
marcofranssen, line 71:
```suggestion
          - {{ printf "%s-%s-webhook" .Release.Namespace (include "spire-controller-manager.fullname" .) }}
```

Maybe use the same style in both places

PR #107, opened 2023-11-16
Fix typo ClusterSPIFFEID for workloadTSelectoremplates

MattiasGees:
I found a small typo in the ClusterSPIFFEID spec and this small change will fix it.


mrsabath (approved):
Good find! LGTM

kfox1111 (approved):
Good catch. Thanks for fixing it! :)

PR #108, opened 2023-11-16
Add flag to install into recommended namespaces

kfox1111:
Allow the user to specify a flag to install all the objects into the recommended namespaces. This makes it significantly easier to install in a production environment without needing external values.


kfox1111:
This prototype has served its purpose. Closing in favor of:
* https://github.com/spiffe/helm-charts-hardened/pull/127

PR #109, opened 2023-11-16
Consolidate all the examples to a common relative path

mrsabath:
We need to decide which format we want to use across all the examples:

1. `helm upgrade --install --namespace spire-server spire charts/spire -f examples/production/values.yaml -f examples/openshift/values.yaml`

1. `helm upgrade --install --namespace spire-server spire ../../charts/spire -f values.yaml -f ../openshift/values.yaml`

Personally, I prefer the case 1.


faisal-memon (approved):
Changes look good to me. Thanks @mrsabath for cleaning this up.

PR #112, opened 2023-11-17
Enable agent to kubelet connection to use hostname

kfox1111:


charts/spire/charts/spire-agent/templates/_helpers.tpl
```diff
@@ -104,3 +104,16 @@ Create the name of the service account to use
 {{- print .Values.socketPath }}
 {{- end }}
 
+{{- define "spire-agent.connect_by_hostname" -}}
```
marcofranssen, line 107:
```suggestion
{{- define "spire-agent.connect-by-hostname" -}}
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -1,4 +1,5 @@
 {{- $configSum := (include (print $.Template.BasePath "/configmap.yaml") . | sha256sum) }}
+{{- $cbh := eq (include "spire-agent.connect_by_hostname" .) "true" }}
```
marcofranssen, line 2:
```suggestion
{{- $cbh := eq (include "spire-agent.connect-by-hostname" .) "true" }}
```

marcofranssen (changes requested):
Small nitpick to use the same naming convention as used by other macros.

marcofranssen (approved):
LGTM 🚀

PR #120, opened 2023-11-29
Fix missing protocol in JWT Issuer 

mrsabath:
The JWT issuer needs a protocol to work correctly


faisal-memon (approved):

kfox1111:
https://github.com/awsdocs/amazon-api-gateway-developer-guide/issues/79

Seems like some clients care, some don't. but the oidc spec says it should be there. So, yeah, I think this could be considered a bug and should be fixed.

kfox1111 (approved):
LGTM

PR #121, opened 2023-11-29
Add a flag to enable recommendations

kfox1111:
Other pr's will build upon this base flag and layout.


kfox1111:
Nows the time to have a discussion on what to name this flag (naming's hard). Originally I had something like:

```
    useRecommended:
      securityContexts: false
      priorityClasses: false
      ...
```

Which makes sense grammatically. Your using the recommended security context, for example.

Then I added 'enabled' to it, so you could enable all the flags at once:
```
    useRecommended:
      enabled: false
      securityContexts: true
      priorityClasses: true
      ...
```
User typically only sets userRecommended.enabled=true

But useRecommended.enabled sounds a little weird.

What do you think? There are probably other ways of doing it....

```
  useRecommendations: false
  useRecommended:
    securityContexts: true
    priorityClasses: true
    ...
```
User typically only sets useRecommendations=true

or
```
recommendations:
  enabled: false
  securityContexts: true
  priorityClasses: true
  ...
```
User typically only sets recommendations.enabled=true

sabre1041:
> > Nows the time to have a discussion on what to name this flag (naming's hard). Originally I had something like:
> > ```
> >     useRecommended:
> >       securityContexts: false
> >       priorityClasses: false
> >       ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which makes sense grammatically. Your using the recommended security context, for example.
> > Then I added 'enabled' to it, so you could enable all the flags at once:
> > ```
> >     useRecommended:
> >       enabled: false
> >       securityContexts: true
> >       priorityClasses: true
> >       ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > User typically only sets userRecommended.enabled=true
> > But useRecommended.enabled sounds a little weird.
> > What do you think? There are probably other ways of doing it....
> > ```
> >   useRecommendations: false
> >   useRecommended:
> >     securityContexts: true
> >     priorityClasses: true
> >     ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > User typically only sets useRecommendations=true
> > or
> > ```
> > recommendations:
> >   enabled: false
> >   securityContexts: true
> >   priorityClasses: true
> >   ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > User typically only sets recommendations.enabled=true
> 
> I personally like the last option the most:
> 
> ```
> recommendations:
>   enabled: false
>   securityContexts: true
>   priorityClasses: true
> ```

I agree as well. Its much simpler

marcofranssen:
what about the following:

```yaml
recommendations:
  enabled: false
  securityContexts: true
  priorityClasses: true
```

This way people can opt in on the recommendations by setting `recommendations.enabled=true` and opt out of specific ones by setting them to false.`

This nicely groups the entire feature in a single yaml block.

kfox1111:
Looks like consensus is for:
recommendations

Updating the pr for that layout.

charts/spire/README.md
```diff
@@ -158,6 +158,7 @@ Now you can interact with the Spire agent socket from your own application. The
...
 | `global.spire.trustDomain`                   | The trust domain for Spire install                                                                                                                                                                                                     | `example.org`     |
 | `global.spire.upstreamServerAddress`         | Set what address to use for the upstream server when using nested spire                                                                                                                                                                | `""`              |
+| `global.spire.recommendations.enabled`       | Enable usage of the recommendations. Default is off.                                                                                                                                                                                   | `false`           |
```
faisal-memon, line 161:
```suggestion
| `global.spire.recommendations.enabled`       | Use recommended values for production deployments. Default is off.                                                                                                                                                                                   | `false`           |
```

faisal-memon: Sorry, put in the wrong spot.

charts/spire/values.yaml
```diff
@@ -21,6 +21,10 @@ global:
     ## @param global.spire.upstreamServerAddress Set what address to use for the upstream server when using nested spire
     upstreamServerAddress: ""
 
+    ## @param global.spire.recommendations.enabled Enable usage of the recommendations. Default is off.
```
faisal-memon, line 24:
```suggestion
    ## @param global.spire.recommendations.enabled Use recommended settings for production deployments. Default is off.
```

charts/spire/README.md
kfox1111: No worries.

faisal-memon (approved):

PR #122, opened 2023-11-29
SELinux support

kfox1111:
Add support to the chart to set the SELinux context to enable a working system. Enable it by default on OpenShift clusters.


charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -141,3 +141,16 @@ restrictedScc:
   name: ""
   ## @param restrictedScc.version Version of the restricted SCC
   version: 2
+
+selinux:
```
faisal-memon, line 145: Should this have an enabled flag?

PR #123, opened 2023-11-29
Tornjak UBI support

kfox1111:
The Tornjak containers now have two different flavors. Vanilla and UBI. Automatically select the UBI image when deploying on OpenShift.


mrsabath (approved):
Thank you @kfox1111 for this change. It simplifies the deployment as it automates the Tornjak image selection. I tested it and it works well. LGTM!

PR #125, opened 2023-11-29
Add recommendation for securityContext and podSecurityContext

kfox1111:


charts/spire/templates/_spire-lib.tpl
```diff
@@ -248,6 +248,48 @@ to merge in values, but spire needs arrays.
...
+
+{{/*
+Note: Values not needed on openshift due to it autoassigning restricted users
+*/}}
+{{- define "spire-lib.podsecuritycontext" }}
+{{- $vals := dict "fsGroupChangePolicy" "OnRootMismatch" }}
+{{- if not (dig "openshift" false .Values.global) }}
+{{- $_ := set $vals "runAsUser" 1000 }}
+{{- $_ := set $vals "runAsGroup" 1000 }}
+{{- $_ := set $vals "fsGroup" 1000 }}
```
faisal-memon, line 287: Is it possible to make this yaml like the defult security context section?

kfox1111: Yes. But would make for more code. As it would be something like:
```
{{- define "spire-lib.default_podsecuritycontext_values" }}
runAsUser: 1000
runAsGroup: 1000
fsGroup: 1000
{{- end }}
```

Then at 285-286,
```
{{- $vals = mergeOverwrite $vals (include "default_podsecuritycontext_values" .) }}
```

So, 6 lines vs 3 lines.

It does look a little more readable though.

If you like that code better though, I'm happy to do it.

kfox1111: Actually, those options are specific to vanilla k8s but don't work on openshift... so maybe:
```
{{- define "spire-lib."default_vanilla_extra_podsecuritycontext_values" }}
```
?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -707,6 +707,8 @@ tornjak:
     # requests:
     #   cpu: 100m
     #   memory: 128Mi
+  ## @param tornjak.securityContext [object] Security Context to use
```
marcofranssen, line 710:
```suggestion

  ## @param tornjak.securityContext [object] Security Context to use
```

Maybe white line in between for readability

charts/spire/templates/_spire-lib.tpl
marcofranssen: Simplicity and readability would be my preference, so contributors don't have to learn to much about expressions when contributing.

kfox1111: Ok. I took a stab at reworking this so that there are two more macro's with the defaults for vanilla k8s and openshift. Its denormalized a bit but should be more clear. Please have a look.

faisal-memon (approved):
LGTM. @marcofranssen would appreciate your final approval.

mrsabath (approved):
LGTM! Thank you @kfox1111

marcofranssen (approved):
LGTM 🚀

PR #131, opened 2023-12-01
Add recommendation for namespacePSS

kfox1111:


charts/spire/templates/spire-server-namespace.yaml
```diff
@@ -1,14 +1,27 @@
-{{- if .Values.global.spire.namespaces.server.create }}
+{{- if or .Values.global.spire.namespaces.create .Values.global.spire.namespaces.server.create }}
+{{-   $labels := dict }}
+{{-   if and (dig "spire" "recommendations" "enabled" false .Values.global) (dig "spire" "recommendations" "namespacePSS" true .Values.global) }}
+{{-     $_ := set $labels "pod-security.kubernetes.io/warn" "restricted" }}
+{{-     $_ := set $labels "pod-security.kubernetes.io/audit" "restricted" }}
+{{-     $_ := set $labels "pod-security.kubernetes.io/enforce" "restricted" }}
```
faisal-memon, line 6: is it possible to break these out into yaml so its easier to read?

kfox1111: Updated with them broken out.

marcofranssen (approved):
LGTM 🚀

marcofranssen:
@faisal-memon this one waits for your comment to be marked as resolved.

charts/spire/values.yaml
```diff
@@ -40,6 +42,8 @@ global:
       registry: ""
 
     namespaces:
+      ## @param global.spire.namespaces.create Set to true to Create all namespaces.
+      create: false
```
faisal-memon, line 46: Do we need the flag at this level and at the nested levels?

kfox1111: On some of my clusters, I will not be able to always be able to create both namespaces, only the spire-system one.  I'm expecting others will be in that same boat. So, being able to turn on namespaces individually is important. But for a lot of folks, you just want it to do it all, so having one flag rather then two saves a fair amount on the minimal required config.

```yaml
global:
  openshift: false # If running on openshift, set to true
  spire:
    recommendations:
      enabled: true
    namespaces:
      create: true
    ingressControllerType: "" # If not openshift, and want to expose services, set to a supported option [ingress-nginx]
    # Update these
    clusterName: example-cluster
    trustDomain: example.org
spire-server:
  ca_subject:
    # Update these
    country: ARPA
    organization: Example
    common_name: example.org
```
~17 lines including comments

vs
``` yaml
global:
  openshift: false # If running on openshift, set to true
  spire:
    recommendations:
      enabled: true
    namespaces:
      server:
        create: true
      system:
        create: true
    ingressControllerType: "" # If not openshift, and want to expose services, set to a supported option [ingress-nginx]
    # Update these
    clusterName: example-cluster
    trustDomain: example.org
spire-server:
  ca_subject:
    # Update these
    country: ARPA
    organization: Example
    common_name: example.org
```
~20 lines including comments

So (20/17 ~= 1.18)... about 18% savings on required minimal config for most users. Without comments, ~20%. So I think the extra flag is worth supporting.

faisal-memon: Ok, thanks for the breakdown. Which one takes precedence?

kfox1111: Its global.spire.namespaces.create "or" global.spire.namespaces.server.create to create the spire-server namespace
Its global.spire.namespaces.create "or" global.spire.namespaces.system.create to create the spire-system namespace
All are defaulted off.

So you can just set:
* global.spire.namespaces.server.create=true to get just spire-server
* global.spire.namespaces.system.create=true to get just spire-system

or if you want both, you can set:
global.spire.namespaces.create=true

(or technically, also setting both global.spire.namespaces.server.create=true and global.spire.namespaces.system.create=true works as expected as well. Just more verbose)

charts/spire/values.yaml
```diff
@@ -40,6 +42,8 @@ global:
       registry: ""
 
     namespaces:
+      ## @param global.spire.namespaces.create Set to true to Create all namespaces.
```
faisal-memon, line 45:
```suggestion
      ## @param global.spire.namespaces.create Set to true to Create all namespaces. If this or either of the namespace specific create flags is set, the namespace will be created.
```

PR #133, opened 2023-12-02
Remove redundant namespace labeling for Openshift

mrsabath:
Since we merged #103 there is no need to label the namespaces for Openshift. This is already done in 
`examples/openshift/openshift-values.yaml`


kfox1111:
Can't do this merge quite yet. Its only enabled in certain configurations that are not documented here yet. It will be automatic in:
https://github.com/spiffe/helm-charts-hardened/pull/131
and documented as in:
https://github.com/spiffe/helm-charts-hardened/blob/recommendations-strict-mode/charts/spire/README.md
which will be split out to its own pr as well in the near future.

faisal-memon:
Thanks @mrsabath. As discussed this change won't be needed with the rest of the production recommendations work.

PR #139, opened 2023-12-08
Update jwt test to work with newer slim images

### LINKED ISSUE #138: openssl-3.2 test breakage
One of our tests is broken when using an image with openssl 3.2 in it. Some details in:
https://github.com/spiffe/helm-charts-hardened/pull/134

drewwells:
fixes #138


charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-keys.yaml
```diff
@@ -29,6 +29,27 @@ spec:
...
+          wget -c $URL -O - | tar --strip-components 2 -C . -xzv "step_${STEP_VERSION}/bin/step"
+          chmod +x /data/step
+          /data/step >/dev/null
+      {{- with .Values.securityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      volumeMounts:
+        - name: data-volume
+          mountPath: /data
```
kfox1111, line 52: Looks like it worked for testing. ❤️

For a solution though, things shouldn't downloaded at runtime...

I think the "smallstep/step-cli:0.25.1" image might do the trick?

no jq command in it but could still use the toolkit image to do the jq work.

drewwells: I'm fine with that. I'll copy the binary out to `/data` and use it in toolkit image below

kfox1111: hmm... that would work, but it may be even simpler to split the last line up:
```
echo $JWT | /data/step crypto jwt verify --key=public.pem --alg=RS256 --subtle
```

to
```
echo $JWT > /data/jwt
```

and then read it in, in an instance of the step container that does
```
cat /data/jwt | /data/step crypto jwt verify --key=public.pem --alg=RS256 --subtle
```

Then you don't need to copy binaries around.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -303,7 +303,21 @@ tests:
...
+    ## @param tests.step.image.pullPolicy The image pull policy
+    ## @param tests.step.image.version This value is deprecated in favor of tag. (Will be removed in a future release)
+    ## @param tests.step.image.tag Overrides the image tag whose default is the chart appVersion
+    ##
+    image:
+      registry: ""
+      repository: smallstep/step-cli
+      pullPolicy: IfNotPresent
+      version: ""
+      tag: latest@sha256:071f01f10ccd3aedeb65362b50a873dacd41b59b4d7e60ad46485e7eceec834e
```
kfox1111, line 320: tag should be: 0.25.1

and, can you please add the following to the "spiffe-oidc-discovery-provider/values.yaml" section under file .github/tests/images.json
```
    {
      "query": "tests.step.image",
      "filter": "^[0-9]\\+\\.[0-9]\\+\\.[0-9]\\+$",
      "sort-flags": ["-t", ".", "-k1,1n", "-k2,2n", "-k3,3n"]
    },
```

This will ensure it receives automatic updates.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/tests/test-keys.yaml
kfox1111: yes. but it would could be less fragile to keep the binaries in their respective containers, as if they stop statically linking the step binary, it would become dependent on the image it was built against. while copying the string from container to container should always work?

drewwells: I didn't see cgo, if it's native to there will not be any shared libraries

kfox1111: Just my feeling that its better to not copy binaries around if you don't have to. I'm not going to block over it though.

kfox1111 (approved):
LGTM

Issue #151, opened 2023-12-15
discovery provider rework


kfox1111:
It has several problems.

Without it enabled, jwt's can't be reliably verified. So it should be enabled by default to ensure jwt functionality of spire works out of the box.

When enabled, it defaults to ACME.

ACME support is broken. There is no way to route the http traffic to the provider to do the handshake to get its certs. It doesn't support dns ACME either. It doesn't persist its certs, putting too much load on LetsEncrypt. It also doesn't share when replicas > 1 putting too much load on LetsEncrypt.

The discovery provider only works if configured with insecureScheme.enabled. Which works well at this point. It is secure in certain configurations so is poorly named, but is hard to guarantee it is in all cases.

Providing your own certs is not supported.

It should be possible to use spire itself to provide the certificates but currently not supported. This probably should be the new default configuration.

PR #152, opened 2023-12-15
SPIFFE OIDC Discovery Provider Rework

kfox1111:
* Adds cert-manager support to issue the Certificate
* Remove older broken ACME support in favor of cert-manager based ACME support.
* Enable bring your own certificate support
* Updates the tests to work with https
* Fine grained spiffeid support

Partially fixes: https://github.com/spiffe/helm-charts-hardened/issues/151


kfox1111:
ACME support is still not working.

kfox1111:
[ ] - Passthrough support on spiffe-oidc-discovery-provider
[ ] - mount bundle on tests and use it to validate certs

kfox1111:
* [ ] - Switch to a released image

mrsabath:
I tested this PR on Openshift and it seems to work with the secure connection as default: 
```
helm-charts-hardened$curl https://oidc-discovery.XXXX.us-east.containers.appdomain.cloud/keys
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
helm-charts-hardened$curl -k  https://oidc-discovery.XXXX.us-east.containers.appdomain.cloud/keys
{
  "keys": [
    {
      "kty": "RSA",
      "kid": "XXXXX",
      "alg": "RS256",
      "n": "XXXX",
      "e": "AQAB"
    },
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -445,13 +445,15 @@ controllerManager:
...
+    defaultSpiffeIDTemplate: spiffe://{{ .TrustDomain }}/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}
     clusterSPIFFEIDs:
       # NOTE you can add multiple uniquely named entries to create multiple ClusterSPIFFEID objects. See example below.
       default:
         ## @param controllerManager.identities.clusterSPIFFEIDs.default.enabled Enable this identity for controller manager
         enabled: true
-        ## @param controllerManager.identities.clusterSPIFFEIDs.default.spiffeIDTemplate Spiffe ID template for identities
-        spiffeIDTemplate: spiffe://{{ .TrustDomain }}/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}
+        ## @param controllerManager.identities.clusterSPIFFEIDs.default.spiffeIDTemplate Set to specify Spiffe ID template for this clusterSPIFFEID
+        spiffeIDTemplate: ""
```
marcofranssen, line 456: This feels a bit like a duplicate from external API point of view. Can't we do the implementation as such it can rely on the original `default.spiffeIDTemplate` field?

kfox1111: Interesting. I did it this way as there isn't anything special about the default object. But it is always there... so we could treat it special like that...

Yeah, let me take a stab at implementing that.

kfox1111: Yeah, that worked out well. Updated.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/deployment.yaml
```diff
@@ -1,3 +1,6 @@
+{{- if and (.Values.enabled) (eq .Values.tlsSecret "") (not .Values.certManager.enabled) (not .Values.insecureScheme.enabled) }}
+{{-   fail "You must specify tlsSecret or certManager.enabled or insecureScheme.enabled" }}
+{{- end }}
```
kfox1111, line 3: Placeholder until https://github.com/spiffe/helm-charts-hardened/pull/163 merges. Then built in spire support takes its spot.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/deployment.yaml
```diff
@@ -47,7 +53,7 @@ spec:
             - containerPort: 8008
               name: healthz
           {{- if not .Values.insecureScheme.enabled }}
-            - containerPort: 443
+            - containerPort: 8443
```
kfox1111, line 56: port switched to an unprivileged port so it can run as nonroot. The service can stay 443 so only the pod needs to change.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -88,6 +88,9 @@ livenessProbe:
 ## @param podAnnotations [object] Pod annotations for Spire OIDC discovery provider
 podAnnotations: {}
 
+## @param tlsSecret Provide your own certificate/key via tls style Kubernetes Secret
+tlsSecret: ""
```
faisal-memon, line 92: I think it would be good to link this, cert-manager, and spiffe-helper as the source of certs for this. Maybe have all 3 under a sub section?

kfox1111: hmm... what would that look like?

maybe something like:
```
tls:
  mode: [spiffe or cert-manager or external or none]
  externalSecret: "" #used if mode==external
  cert-manager: ...... stuff from below?
```

where mode=none replaces the insecureScheme.enabled flag?

faisal-memon: I think we can do it without the enum:
```
tls:
  spire:
    enabled: true
    ....

  externalSecret:
    enabled: false
    secretName: my-secret

  cert-manager:
     enabled: true
```
And print an error if more than one is set. What do you think?

kfox1111: Updated.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -88,10 +88,46 @@ livenessProbe:
 ## @param podAnnotations [object] Pod annotations for Spire OIDC discovery provider
 podAnnotations: {}
 
-insecureScheme:
-  ## @param insecureScheme.enabled Flag to enable insecure schema
-  enabled: false
+tls:
```
faisal-memon, line 91:
```suggestion
# Select from one of the options below to be the source of certificates for OICD Discovery Provider.
# If none are enabled, connections will not be TLS encrypted.
tls:
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -88,10 +88,46 @@ livenessProbe:
...
+  spire:
+    ## @param tls.spire.enabled Use spire to secure the oidc-discovery-provider
+    enabled: true
+
+  externalSecret:
+    ## @param tls.externalSecret.enabled Provide your own certificate/key via tls style Kubernetes Secret
+    enabled: false
+    ## @param tls.externalSecret.secretName Specify which secret to use
```
faisal-memon, line 99:
```suggestion
    ## @param tls.externalSecret.secretName Specify which Secret to use
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -88,10 +88,46 @@ livenessProbe:
...
+    enabled: false
+    issuer:
+      ## @param tls.certManager.issuer.create Create an issuer to use to issue the certificate
+      create: true
+      acme:
+        ## @param tls.certManager.issuer.acme.email Must be set in order to register with LetsEncrypt. By setting, you agree to their Terms of Service
+        email: ""
+        ## @param tls.certManager.issuer.acme.server Server to use to get certificate. Defaults to LetsEncrypt
+        server: https://acme-v02.api.letsencrypt.org/directory
+        # Testing server: https://acme-staging-v02.api.letsencrypt.org/directory
```
faisal-memon, line 113: What is Testing server for?

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -88,10 +88,46 @@ livenessProbe:
...
+      dnsNames: []
+      ## @param tls.certManager.certificate.issuerRef.group If you are using an external plugin, specify the group for it here
+      ## @param tls.certManager.certificate.issuerRef.kind Kind of the issuer reference. Override if you want to use a ClusterIssuer
+      ## @param tls.certManager.certificate.issuerRef.name Name of the issuer to use. If unset, it will use the name of the built in issuer
+      issuerRef:
+        group: ""
+        kind: Issuer
+        name: ""
+
+insecureScheme:
```
faisal-memon, line 130: Why is nginx here under insecureScheme?

charts/spire/charts/spiffe-oidc-discovery-provider/templates/_helpers.tpl
```diff
@@ -91,3 +91,21 @@ Create the name of the service account to use
...
+{{-     $tlsCount = add $tlsCount 1 }}
+{{-   end }}
+{{-   if and .Values.enabled .Values.tls.externalSecret.enabled }}
+{{-     $tlsCount = add $tlsCount 1 }}
+{{-   end }}
+{{-   if and .Values.enabled .Values.tls.certManager.enabled }}
+{{-     $tlsCount = add $tlsCount 1 }}
+{{-   end }}
+{{-   if gt $tlsCount 1 }}
+{{-     fail "You can only have one tls configuration enabled" }}
```
faisal-memon, line 108:
```suggestion
{{-     fail "You can only have one TLS configuration enabled" }}
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: historical. Want me to move that up to the root?

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -88,10 +88,46 @@ livenessProbe:
...
+        ## @param tls.certManager.issuer.acme.server Server to use to get certificate. Defaults to LetsEncrypt
+        server: https://acme-v02.api.letsencrypt.org/directory
+        # Testing server: https://acme-staging-v02.api.letsencrypt.org/directory
+        ## @param tls.certManager.issuer.acme.solvers [object] Configure the issuer solvers. Defaults to http01 via ingress.
+        solvers: {}
+        # - http01:
+        #     ingress:
+        #       ingressClassName: nginx
+    certificate:
+      ## @param tls.certManager.certificate.dnsNames Override the dnsNames on the certificate request. Defaults to the same settings as Ingress
```
faisal-memon, line 122: Is there value in making this function like additional dns names?

charts/spire/charts/spiffe-oidc-discovery-provider/templates/ingress.yaml
```diff
@@ -1,13 +1,36 @@
 {{- if .Values.ingress.enabled -}}
+{{- $tlsCount := include "spiffe-oidc-discovery-provider.tls-count" . | int }}
+{{- $port := 443 }}
```
faisal-memon, line 3: Should this port be hardcoded?

charts/spire/charts/spiffe-oidc-discovery-provider/templates/_helpers.tpl
```diff
@@ -91,3 +91,21 @@ Create the name of the service account to use
 {{- define "spiffe-oidc-discovery-provider.workload-api-socket-path" -}}
 {{- printf "/spiffe-workload-api/%s" .Values.agentSocketName }}
 {{- end }}
+
+{{- define "spiffe-oidc-discovery-provider.tls-count" -}}
+{{-   $tlsCount := 0 }}
+{{-   if and .Values.enabled .Values.tls.spire.enabled }}
+{{-     fail "Built in spire support is not yet supported." }}
```
faisal-memon, line 98:
```suggestion
{{-     fail "Built in SPIRE support is not yet supported." }}
```

charts/spire/charts/spiffe-oidc-discovery-provider/templates/_helpers.tpl
```diff
@@ -91,3 +91,21 @@ Create the name of the service account to use
 {{- define "spiffe-oidc-discovery-provider.workload-api-socket-path" -}}
 {{- printf "/spiffe-workload-api/%s" .Values.agentSocketName }}
 {{- end }}
+
+{{- define "spiffe-oidc-discovery-provider.tls-count" -}}
```
faisal-memon, line 95: Is it possible to structure this so this check to runs once pre deploy? And all the templates that need to know if tls is enabled can have a simpler template `is-tls` thats just an or of the enabled flags.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
faisal-memon: is it used when tls is enabled?

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
faisal-memon: Thanks @kfox1111 for refactoring. I think its easier to follow now.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: letsencrypt has two different instances. production and their testing instance. Production has very strict rules and if you misconfigure too badly, you can ban yourself. Testing is useful for some folks to make sure everything handshakes properly before switching to production. Though with cert-manager, I've never seen it misconfigured badly enough to ban on their prod server.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/_helpers.tpl
kfox1111: Unfortunately not.

kfox1111: Well..... Maybe....Let me take a stab at that...

charts/spire/charts/spiffe-oidc-discovery-provider/templates/ingress.yaml
kfox1111: Found an example of tornjak choosing between two ports... reimplementing that way to match...

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: no. only when there is no tls.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/_helpers.tpl
kfox1111: Updated.

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
kfox1111: Letsencrypt in some configurations supports wildcard certificates. So instead of using the default certificate request, you may want to override it all and just set
```
dnsNames:
- *.example.com
```

charts/spire/charts/spiffe-oidc-discovery-provider/templates/ingress.yaml
kfox1111: Done

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -88,10 +91,48 @@ livenessProbe:
...
+      ## @param tls.certManager.certificate.issuerRef.kind Kind of the issuer reference. Override if you want to use a ClusterIssuer
+      ## @param tls.certManager.certificate.issuerRef.name Name of the issuer to use. If unset, it will use the name of the built in issuer
+      issuerRef:
+        group: ""
+        kind: Issuer
+        name: ""
+
+insecureScheme:
   nginx:
     ## @param insecureScheme.nginx.image.registry The OCI registry to pull the image from
```
faisal-memon, line 137:
```suggestion
    ## @param insecureScheme.nginx.image.registry The OCI registry to pull the image from. Only used when TLS is disabled.
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
```diff
@@ -88,10 +91,48 @@ livenessProbe:
...
     ## @param insecureScheme.nginx.image.repository The repository within the registry
```
faisal-memon, line 138:
```suggestion
    ## @param insecureScheme.nginx.image.repository The repository within the registry. Only used when TLS is disabled.
```

charts/spire/charts/spiffe-oidc-discovery-provider/values.yaml
marcofranssen: staging is very useful when testing infrastructure automation. E.g. create destroy workflows during CI.

faisal-memon (approved):
Thanks @kfox1111 for making the requested changes.

PR #153, opened 2023-12-15
Update HorizontalPodAutoscaler API to autoscaling/v2

### LINKED ISSUE #140: Update HorizontalPodAutoscaler API to autoscaling/v2
autoscaling/v2beta1 was removed in Kubernetes 1.25: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25 and autoscaling/v2 is available since Kubernetes 1.23.
We should change the SPIRE server and OIDC Discovery Provider charts to use autoscaling/v2 for the HPA.

I am more than happy to open a PR for this!

jer8me:
This PR is to close https://github.com/spiffe/helm-charts-hardened/issues/140:
  - autoscaling/v2beta1 is deprecated since Kubernetes 1.25 (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25)
  - for SPIRE server, set the scaleTargetRef kind to StatefulSet
  - reorder metrics (to avoid issues like https://github.com/argoproj/argo-cd/issues/1079)

closes #140


charts/spire/charts/spiffe-oidc-discovery-provider/templates/hpa.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2
```
marcofranssen, line 2: Can we check for the kube verison here that introduced this version, so it doesn't break for people running older clusters?

charts/spire/charts/spiffe-oidc-discovery-provider/templates/hpa.yaml
```diff
@@ -14,16 +14,20 @@ spec:
...
-    {{- if .Values.autoscaling.targetMemoryUtilizationPercentage }}
+    {{- if .Values.autoscaling.targetCPUUtilizationPercentage }}
     - type: Resource
       resource:
-        name: memory
-        targetAverageUtilization: {{ .Values.autoscaling.targetMemoryUtilizationPercentage }}
+        name: cpu
+        target:
+          type: Utilization
+          averageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
```
marcofranssen, line 31: Any specific reason to switch the order here? 

Can we leave the original order to reduce the diff in this PR?

charts/spire/charts/spire-server/templates/hpa.yaml
```diff
@@ -9,24 +9,28 @@ metadata:
 spec:
   scaleTargetRef:
     apiVersion: apps/v1
-    kind: Deployment
+    kind: StatefulSet
```
marcofranssen, line 12: Ouch that is a bug for a while. Good catch.

charts/spire/charts/spire-server/templates/hpa.yaml
```diff
@@ -9,24 +9,28 @@ metadata:
...
-    {{- if .Values.autoscaling.targetMemoryUtilizationPercentage }}
+    {{- if .Values.autoscaling.targetCPUUtilizationPercentage }}
     - type: Resource
       resource:
-        name: memory
-        targetAverageUtilization: {{ .Values.autoscaling.targetMemoryUtilizationPercentage }}
+        name: cpu
+        target:
+          type: Utilization
+          averageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
```
marcofranssen, line 34: Is it required to change the order here?

marcofranssen (changes requested):
Thanks for your contribution. Small suggestion on the apiVersion change.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/hpa.yaml
jer8me: With the existing order, Argo-CD thinks that there is a diff because the order ends up being modified. The Argo-CD issue I mentioned in the PR suggests reordering the metrics as a possible workaround (in my case I do not own the argo configuration): https://github.com/argoproj/argo-cd/issues/1079#issuecomment-463849329

charts/spire/charts/spiffe-oidc-discovery-provider/templates/hpa.yaml
jer8me: Sounds good, will add a kube version check.

jer8me: Since this chart is for k8s 1.21 or above, I think we should at least upgrade to autoscaling/v2beta2 for people running older clusters. autoscaling/v2beta1 was deprecated in k8s 1.19 and has a different API (resource.targetAverageUtilization vs resource.target.type/resource.target.averageUtilization).

So I am thinking about checking .Capabilities.APIVersions to see if autoscaling/v2 is available and if not, go with autoscaling/v2beta2.

charts/spire/charts/spire-server/templates/hpa.yaml
kfox1111: hmm.....

If this is the case, the chart on k8s 1.21 / 1.22 has never worked with hpa.... they introduced apps/v2 in k8s 1.23.
Maybe we should just support v2 on this object since its very unlikely to affect anyone at 1.21 (Won't affect my 1.21 cluster)

charts/spire/charts/spiffe-oidc-discovery-provider/templates/hpa.yaml
marcofranssen: Sounds good

charts/spire/charts/spire-server/templates/hpa.yaml
marcofranssen: Didn't we switch spire-server a while ago to a Statefulset. Originally it was a Deployment.

marcofranssen:
@jer8me could you pull in the changes from upstream/main?

charts/spire/charts/spire-server/templates/hpa.yaml
kfox1111: It could have easily been a Deployment in the very beginning, yeah.

jer8me:
I implemented the autoscaling API logic and pulled main in my branch. Waiting on some internal reviews before I can push my changes here.

charts/spire/charts/spire-server/templates/hpa.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2
```
kfox1111, line 2: A thought...

Since this file came about, we added a check here:
https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-server/templates/statefulset.yaml#L19-L21

to ensure scaleup doesn't happen when the database type doesn't support scaling. Could you add a similar check (maybe just fail outright if sqlite & autoscaling enabled) between line 1 and 2 please?

jer8me: There is an existing check in hpa.yaml: https://github.com/spiffe/helm-charts-hardened/blob/spire-0.16.0/charts/spire/charts/spire-server/templates/hpa.yaml#L15
It should fail if sqlite is selected and the number of replicas is greater than 1.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/hpa.yaml
kfox1111, line 31: (Note to other reviewers) I double checked the target.x style is correct for v2beta2.

marcofranssen (approved):
LGTM 🚀 @jer8me thanks a lot for this contribution 🎉

PR #166, opened 2023-12-27
Add basic CSI driver for the spiffe-helper

kfox1111:
This adds a prototype level helm chart for a virtual csi driver that makes it significantly easier to integrate existing software running on Kubernetes with SPIRE.


.github/tests/charts.json
```diff
@@ -23,5 +23,10 @@
     "name": "postgresql",
     "repo": "https://charts.bitnami.com/bitnami",
     "version": "14.0.1"
+  },
+  {
+    "name": "kyverno",
+    "repo": "https://kyverno.github.io/kyverno",
+    "version": "3.1.1"
```
edwbuck, line 30: Let's not slip in a policy engine as an out-of-scope "extra" on a commit intended to add a CSI driver.

.github/tests/pre-install.sh
```diff
@@ -47,3 +47,8 @@ kubectl wait --namespace ingress-nginx --for=condition=ready --timeout 60s pod -
   --namespace postgresql \
   --values "${DEPS}/postgresql.yaml" \
   --wait
+
+# kyverno
+"${helm_install[@]}" kyverno kyverno --version "$VERSION_KYVERNO" --repo "$HELM_REPO_KYVERNO" \
```
edwbuck, line 52: Let's not slip in a policy engine as an out-of-scope "extra" on a commit intended to add a CSI driver.

charts/spiffe-helper-csi-driver/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Source Code
+
+* <https://github.com/spiffe/helm-charts/tree/main/charts/spiffe-helpe-csi-driver>
+
+## Prereqs:
+
+Your cluster needs to have Kyverno installed. You can do that by running something like the following:
+
+```
+helm upgrade --install --create-namespace kyverno kyverno -n kyverno --repo https://kyverno.github.io/kyverno/ --version 3.1.1
```
edwbuck, line 27: Let's not slip in a policy engine as an out-of-scope "extra" on a commit intended to add a CSI driver.

charts/spiffe-helper-csi-driver/templates/_spire-lib.tpl
```diff
@@ -0,0 +1,26 @@
+{{- define "spire-lib.registry" }}
+{{- if ne (len (dig "spire" "image" "registry" "" .global)) 0 }}
```
edwbuck, line 2: Please use the (((x).y).z) pattern of protecting against missing elements instead of dig.

charts/spiffe-helper-csi-driver/templates/_spire-lib.tpl
```diff
@@ -0,0 +1,26 @@
...
+{{- print .global.spire.image.registry "/"}}
```
edwbuck, line 3: Add indentation.  It helps maintenance.

charts/spiffe-helper-csi-driver/templates/_spire-lib.tpl
```diff
@@ -0,0 +1,26 @@
...
+{{- else if ne (len (.image.registry)) 0 }}
+{{- print .image.registry "/"}}
+{{- end }}
+{{- end }}
+
+{{- define "spire-lib.image" -}}
+{{- $registry := include "spire-lib.registry" . }}
+{{- $repo := .image.repository }}
+{{- $tag := (default .image.tag .image.version) | toString }}
```
edwbuck, line 12: If both are unset, is it valid to have a $tag of ""?

charts/spiffe-helper-csi-driver/templates/_spire-lib.tpl
```diff
@@ -0,0 +1,26 @@
...
+{{- if ne (len (dig "spire" "image" "registry" "" .global)) 0 }}
+{{- print .global.spire.image.registry "/"}}
+{{- else if ne (len (.image.registry)) 0 }}
+{{- print .image.registry "/"}}
+{{- end }}
+{{- end }}
+
+{{- define "spire-lib.image" -}}
+{{- $registry := include "spire-lib.registry" . }}
+{{- $repo := .image.repository }}
```
edwbuck, line 11: No handling for unset values here.

charts/spiffe-helper-csi-driver/templates/_spire-lib.tpl
```diff
@@ -0,0 +1,26 @@
...
+{{- $tag := (default .image.tag .image.version) | toString }}
+{{- if eq (substr 0 7 $tag) "sha256:" }}
```
edwbuck, line 13: Use hasPrefix instead of the combined "if eq (substr ..."

charts/spiffe-helper-csi-driver/templates/_spire-lib.tpl
```diff
@@ -0,0 +1,26 @@
...
+{{- printf "%s/%s@%s" $registry $repo $tag }}
+{{- else if .appVersion }}
+{{- $appVersion := .appVersion }}
+{{- if and (hasKey . "ubi") (dig "openshift" false .global) }}
+{{- $appVersion = printf "ubi-%s" $appVersion }}
+{{- end }}
+{{- printf "%s%s:%s" $registry $repo (default $appVersion $tag) }}
```
edwbuck, line 20: This is printf'ing things that were composed by printf'ing above.  Might be better to consolidate, such that we only have one printf involved in each "condition"

charts/spiffe-helper-csi-driver/templates/clusterpolicy.yaml
```diff
@@ -0,0 +1,4 @@
+{{- $cp := .Files.Get "files/clusterpolicy.yaml" }}
+{{- $spiffe_helper_image := include "spire-lib.image" (dict "image" .Values.spiffeHelper.image "global" .Values.global) }}
+{{- $busybox_image := include "spire-lib.image" (dict "image" .Values.busybox.image "global" .Values.global) }}
+{{- $cp | replace "@SPIFFE_HELPER_IMAGE@" $spiffe_helper_image | replace "@BUSYBOX_IMAGE@" $busybox_image }}
```
edwbuck, line 4: Why do we have to encode templating operations in a template?

charts/spiffe-helper-csi-driver/values.yaml
```diff
@@ -0,0 +1,22 @@
+## @skip global
+global: {}
+
+## @param spiffeHelper.image.registry The OCI registry to pull the image from
```
edwbuck, line 4: Might want to clarify "spiffe helper image"

edwbuck (commented):
A good start.  Needs a little cleanup here and there.  I'm a bit confused why we need to apply values to clusterpolicy.yaml in code, when it likely should be handled by templating.

.github/tests/charts.json
kfox1111: The whole hook is implemented as a kyverno mutating policy. Its needed to make it work at all.

Totally open to it needing its own golang based webhook implementation in the future, along with associated git repositories, a container, etc, to make it not depend on kyverno, but was looking at getting an implementation of the concept working in as minimal amount of time possible, and that lead to using an existing webhook engine.

charts/spiffe-helper-csi-driver/templates/_spire-lib.tpl
kfox1111: Anything _spire-lib.tpl related was just copy/pasted here since at the time the pr was written, spire-lib was not broken out to its own chart. This is fixed now in trunk, and once the pr is rebased, all traces of the _spire-lib.tpl file will go away in this pr.

Its probably worth noting these changes to be made to charts/spire/charts/spire-lib in a seperate issue so they can be made to the main file still containing the recommended changes. Could you please file an issue?

charts/spiffe-helper-csi-driver/templates/clusterpolicy.yaml
kfox1111: There are two template engines at play here. Helms, and Kyverno's. In order to make it easy to helm install and set the helper image to be used, it needs to run that part through helm's template engine.

charts/spiffe-helper-csi-driver/values.yaml
kfox1111:
```suggestion
## @param spiffeHelper.image.registry The OCI registry to pull the spiffe helper image from
```

charts/spiffe-helper-csi-driver/templates/clusterpolicy.yaml
edwbuck: The expansion of the charts to include Kyverno is something that probably should be discussed with a majority of the maintainers.  I understand that it's part of this implementation, but its inclusion should be done with consensus, and at least some commentary on what the alternatives might be.

.github/tests/charts.json
edwbuck: The expansion of the charts to include Kyverno is something that probably should be discussed with a majority of the maintainers.  I understand that it's part of this implementation, but its inclusion should be done with consensus, and at least some commentary on what the alternatives might be.

kfox1111: Completely agree.

kfox1111: I 100% believe Kyverno is **NOT** the right solution in the long run.

The right place for it though is not clear. There are several potential places
1. kubernetes is working on gaining cel support for MutatingAdmissionPolicy. This would be the very best place to put it, I think. provided it supports all the needed functionality (unclear). But it isn't supported in k8s yet, let alone in all versions of k8s we support.
2. It could be added to spire-controller-manager. Another webhook wouldn't be required then.
3. Its own standalone repo/webhook - needs ssc support

---

I also don't think we should block users playing around with the concept and getting the api correct so that when we do identify a better way of implementing that api, we can switch out the kyverno based implementation with a better one without affecting users of it.

charts/spiffe-helper-csi-driver/templates/clusterpolicy.yaml
kfox1111: Duplicated thread from above. closing this one in favor of continuing discussion above

edwbuck:
partially blocked on https://github.com/spiffe/spiffe-helper/pull/118

We need a deeper discussion on whether to bring in Kyverno

kfox1111:
Ideally, the webhook will be implemented in go or kubernetes mutating cel. A home for that needs to be established.

The api, tests, and chart would be unchanged and should be good to go.

charts/spiffe-helper-csi-driver/README.md
```diff
@@ -0,0 +1,56 @@
+# spire-helper-csi-driver
+
+![Version: 0.1.0](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.7.2](https://img.shields.io/badge/AppVersion-1.7.2-informational?style=flat-square)
+
+A Helm chart to install the SPIFFE HELPER CSI Driver.
```
kfox1111, line 5:
```suggestion
A Prototype Helm chart to install the SPIFFE HELPER CSI Driver. It is useful only for testing at this point. Please do not use in production.
```

.github/tests/charts.json
marcofranssen: For what it is worth, in our own clusters we manage Kyverno ourself, meaning I probably need a way to disable it here, for the remainder I understand the reason why you are adding it now.

Probably it should be something managed in the controller itself to have the webhook implemented there.

kfox1111: It makes use of Kyverno via a policy in the chart, but it doesn't install Kyvverno other then in the tests, to test the policy.

Longer term, it will be switched to https://github.com/kubernetes/kubernetes/pull/127134 once it is a thing.

kfox1111:
There is a work in progress spiffefs driver that we will probably replace this with. PR's to come soon.

PR #184, opened 2024-01-12
Fix agent daemonset format

mrsabath:
Very minor formatting bug, that causes formatting errors when OpenShift enabled.


kfox1111 (approved):
Oh, good catch. Thanks!

PR #189, opened 2024-01-18
Update the CRs to enable multiple instance nesting without naming conflicts

kfox1111:


marcofranssen (commented):
Could you add in the description of this PR how this can/should be utilized?

kfox1111:
When combining the three releases as described above and in this diagram: https://github.com/kfox1111/spiffe.io/blob/helm-charts-hardened-nested/static/img/spire-helm-charts-hardened/singlehardened.png, there are naming conflicts between each instance as Release.Name is the same in each case. This PR makes them overridable for each child instance.

marcofranssen (approved):
LGTM 🚀

PR #197, opened 2024-01-20
Remove CRD that isn't really a CRD

kfox1111:


kfox1111:
Confirmed with @azdagron that this shouldn't be loaded as a CRD into Kubernetes.

faisal-memon (approved):

PR #205, opened 2024-01-24
Update spire-server with default container annotation.

cccsss01:
Adding default container annotation to the stateful set


marcofranssen (approved):
LGTM 🚀

PR #206, opened 2024-01-24
Adding dependency StorageClass

cccsss01:
Adding dependency storage class to production README.md I forgot this was required.


examples/production/README.md
```diff
@@ -1,5 +1,10 @@
 # Recommended production setup
 
+Dependencies
+```
+StorageClass
+```
```
marcofranssen, line 6:
```suggestion

- StorageClass
```

kfox1111:
I think this pr's good. But raises a further question. With the new docs in the chart at  https://artifacthub.io/packages/helm/spiffe/spire#production

and similar docs lined up for the spiffe website itself, should we change this file to just point at the other docs? Its kind of out of sync with those recommendations.

cccsss01:
Sounds good to me.

cccsss01:
On second thought I think i'd like this open, and likely add a disclaimer stating to check other pages.

cccsss01:
How is artifacthub documentation version controlled?

kfox1111:
The artifacthub page is just displaying the README.md out of the charts. So in some ways its more relevant then whats in the git repo, as it is the README.md associated with the particular version of the chart.

kfox1111:
Filed https://github.com/spiffe/helm-charts-hardened/issues/217 for more discussion

faisal-memon:
@cccsss01 thanks for taking the time for this. Our plan is to delete the `examples/` folder as its fallen out of date so we dont need to update this readme anymore.

kfox1111:
examples/production specifically.

PR #207, opened 2024-01-24
Always upgrade test from the previous major release

kfox1111:
Previously the upgrade test had a hard coded version to try to upgrade from. This PR updates it so that it automatically detects the newest major release version in the series but uses the start of the series. So for example, test upgrading from 0.17.0 rather then upgrading from 0.17.1.

### DIFF
--- a/examples/production/run-tests.sh
+++ b/examples/production/run-tests.sh
@@ -26,7 +26,7 @@ for i in "$@"; do
     -u)
       if [[ -z "$UPGRADE_VERSION" ]]; then
         echo "Failed to detect previous version."
-	exit 1
+        exit 1
       fi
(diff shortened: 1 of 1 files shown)


examples/production/run-tests.sh
```diff
@@ -24,6 +24,10 @@ CLEANUP=1
 for i in "$@"; do
   case $i in
     -u)
+      if [[ -z "$UPGRADE_VERSION" ]]; then
+        echo "Failed to detect previous version."
+	exit 1
```
marcofranssen, line 29:
```suggestion
        echo "Failed to detect previous version."
        exit 1
```

marcofranssen (approved):
LGTM 🚀

PR #210, opened 2024-01-25
Reduce spire-agent readinessProbe to allow faster rollouts

marcofranssen:
Overall I see the daemonsets on my nodes be ready in approx 5 to 7 seconds.
Reducing the initial delay allows them to be marked as ready within 10 seconds. In cases it takes longer the next probe will be in 30 seconds so max initial waiting time is 40 seconds to be marked as ready compared to 75 seconds at this point of time. Furhtermore the 30 seconds probes also allow more quick detection of failures.

This also speeds up helm upgrades as the rollout restart of daemonsets go faster.


kfox1111:
+1 to lowering the readiness check values.

I'm not as sure about the liveness ones though. liveness checks that are too low/quick can cause k8s to prematurely shoot the pods when systems are a bit overloaded. They also can be problematic on slower systems like minikube/rasperry pi's. Though that is why they are overridable.

What do you think? Should we do just the readiness check values and leave the liveness ones the way they were? I think that would still have the affect your after. But if its working reliably in low resource environments, I'd be ok with both.

faisal-memon:
Im not familiar with these settings so I cant comment on whether the ones here are good or not. If there is agreement on the readiness check values then can we move forward with merging those and discuss the liveness settings separately?

On a related note, we recently changed the value of `maxUnavailable` to 50% to speed up re rolling daemonsets. Perhaps we can add a configurable for that?

marcofranssen:
Ok will update to only change the readiness for now. Will also add a PR for the updateStrategy maxSurge and such so more pods can reboot at the same time.

kfox1111 (approved):
LGTM. Thanks! :)

PR #211, opened 2024-01-25
Add direct tpm support for spire-server

kfox1111:


kfox1111:
Tested with real hardware. Its good to go.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -684,6 +684,25 @@ nodeAttestor:
   joinToken:
     ## @param nodeAttestor.joinToken.enabled Enable the join_token nodeattestor
     enabled: false
+  tpmDirect:
+    ## @param nodeAttestor.tpmDirect.enabled Enable the direct tpm nodeattestor
```
faisal-memon, line 688:
```suggestion
    ## @param nodeAttestor.tpmDirect.enabled Enable the direct TPM node attestor, a 3rd party plugin by Boxboat.
```

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -191,6 +223,9 @@ spec:
             {{- if gt (len .Values.extraVolumeMounts) 0 }}
             {{- toYaml .Values.extraVolumeMounts | nindent 12 }}
             {{- end }}
+            - name: server-tmp
+              mountPath: /tmp
+              readOnly: false
```
faisal-memon, line 228: What is this used for?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -684,6 +684,25 @@ nodeAttestor:
...
+    ## @param nodeAttestor.tpmDirect.image.pullPolicy The image pull policy
+    ## @param nodeAttestor.tpmDirect.image.tag Overrides the image tag
+    ##
+    image:
+      registry: docker.io
+      repository: boxboat/spire-tpm-plugin-tpm-attestor-server
+      pullPolicy: IfNotPresent
+      tag: "v1.7.0"
+    ## @param nodeAttestor.tpmDirect.checksum The sha256 checksum of the plugin binary
+    checksum: 6dc74979eb41ecbd45cd047592c24e1830f9c52fff40441d69870e8927def13f
```
faisal-memon, line 701: Is this something that needs to be calculated by the user?

kfox1111: SPIRE's config file needs a checksum to verify the plugin has not been tempered with. Since SPIRE itself requires the user to set it, I'm following that convention here. The user can calculate it, but its a little tedious to do so. We set the correct checksum for the image we're recommending here so the user doesn't have to. Ideally the checksum would be provided by the provider of the image. I'll have a chat with them.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -53,8 +53,27 @@ spec:
...
+        {{- if .Values.nodeAttestor.tpmDirect.enabled }}
+        - name: init-tpm-direct
+          securityContext:
+            {{- include "spire-lib.securitycontext" . | nindent 12 }}
+          image: {{ template "spire-lib.image" (dict "appVersion" $.Chart.AppVersion "image" .Values.nodeAttestor.tpmDirect.image "global" .Values.global) }}
+          command:
+            - sh
+            - -ec
+            - |
+              cp -a /app/tpm_attestor_server /tpm
```
kfox1111, line 67: Make the path/filename for the tpm plugin configurable here, as it could be different in different images.

charts/spire/charts/spire-server/values.yaml
kfox1111: https://github.com/boxboat/spire-tpm-plugin/pull/12 for the checksums on their releases to make it into their release notes.

kfox1111: The plugins will have checksums on all releases now. see the text files here: https://github.com/boxboat/spire-tpm-plugin/releases on the newest releases.

charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -53,8 +53,27 @@ spec:
...
+        {{- if .Values.nodeAttestor.tpmDirect.enabled }}
+        - name: init-tpm-direct
+          securityContext:
+            {{- include "spire-lib.securitycontext" . | nindent 12 }}
+          image: {{ template "spire-lib.image" (dict "appVersion" $.Chart.AppVersion "image" .Values.nodeAttestor.tpmDirect.image "global" .Values.global) }}
+          command:
+            - sh
+            - -ec
+            - |
+              cp -a {{ .Values.nodeAttestor.tpmDirect.pluginPath }} /tpm/tpm_attestor_server
```
faisal-memon, line 67: Why is this step needed?

kfox1111: Plugins currently have to be forked off of the spire-server/agent process so must be available within the spire-server/agent container. This makes it available for executation.

faisal-memon: Thanks for the explanation. Can you add that as a comment to the code?

faisal-memon (approved):

charts/spire/charts/spire-server/templates/statefulset.yaml
faisal-memon: Thanks for adding

PR #212, opened 2024-01-26
Configurable daemonsets updateStrategy

marcofranssen:
This allows more deamonset pods to be rotated in the same time. This speeds up the updates.


charts/spire/charts/spiffe-csi-driver/templates/daemonset.yaml
```diff
@@ -9,6 +9,9 @@ spec:
   selector:
     matchLabels:
       {{- include "spiffe-csi-driver.selectorLabels" . | nindent 6 }}
+  {{- with .Values.updateStrategy }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   updateStrategy:
     type: RollingUpdate
```
kfox1111, line 16: Looks like there is a problem here.

charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -37,6 +37,15 @@ healthChecks:
   ## @param healthChecks.port The healthcheck port for spiffe-csi-driver
   port: 9809
 
+## @param updateStrategy.type The update strategy to use to replace existing DaemonSet pods with new pods
+## @param updateStrategy.rollingUpdate.maxUnavailable Max unavailable pods during update
+## @param updateStrategy.rollingUpdate.maxSurge Max pods that simultatiously start the update
+updateStrategy:
+  type: RollingUpdate
```
faisal-memon, line 44: Is the plan to make this an enum of the available update strategies?

marcofranssen: personally I would never go with another strategy. It is in the values because I used the toYaml approach, maybe we should remove the option to change it 🤔

kfox1111: the other option at the moment is 'OnDelete', which I think has some advantages too. Being able to keep things more stable by upgrading the daemonset but not the workload until your node is drained and in a downtime before deleting / upgrading the agent.

marcofranssen: @faisal-memon there is no concept of enums out of the box in yaml/helm. I don't have the time at this point in time to build our own validation using macros.

marcofranssen: @kfox1111 should we go with OnDelete as a more safe default?

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -108,6 +108,15 @@ healthChecks:
...
   port: 9982
 
+## @param updateStrategy.type The update strategy to use to replace existing DaemonSet pods with new pods
+## @param updateStrategy.rollingUpdate.maxUnavailable Max unavailable pods during update
+## @param updateStrategy.rollingUpdate.maxSurge Max pods that simultatiously start the update
+updateStrategy:
+  type: RollingUpdate
+  rollingUpdate:
+    maxUnavailable: 1
+    maxSurge: 0
```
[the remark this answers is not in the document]
marcofranssen: Aaah I misunderstood, yeah in that case we should maybe hardcode the value and do some macro to merge the values with that hardcoded structure

kfox1111:
Made an attempt at implementing the discussed changes.

faisal-memon (approved):

drewwells:
Hero, waiting on some really boring deploys with maxUnavailable: 1 right now. Thanks for making the future look brighter

image

PR #216, opened 2024-01-26
Add direct tpm support for spire-agent

kfox1111:


charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -56,11 +56,27 @@ agent:
...
     k8s_psat:
       plugin_data:
         cluster: {{ include "spire-lib.cluster-name" . | quote }}
+  {{- $nodeAttestorUsed = add1 $nodeAttestorUsed }}
+  {{- end }}
+  {{- with .Values.nodeAttestor.tpmDirect }}
+  {{- if eq (.enabled | toString) "true" }}
+    tpm:
+      plugin_cmd: "/tpm/tpm_attestor_agent"
+      plugin_checksum: {{ .checksum }}
```
kfox1111, line 72:
```suggestion
      plugin_checksum: {{ .plugin.checksum }}
```

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -56,11 +56,27 @@ agent:
   {{- end }}
   {{- end }}
 
+{{- $nodeAttestorUsed := 0 }}
```
kfox1111, line 59:
```suggestion
{{- $nodeAttestorUsed := add (len nodeAttestor) (len .Values.unsupportedBuiltInPlugins.nodeAttestor) }}
```

examples/tpm-direct/values.yaml
```diff
@@ -0,0 +1,14 @@
...
+      enabled: false
+    tpmDirect:
+      enabled: true
+
+spire-server:
+  controllerManager:
+    # K8s labels have a 63 character limit. TPM hashes are 64 chars. So you need to label the node with two labels with half of the tpm's hash each. The 'node-restriction.kubernetes.io/' prefix is so that the
+    # nodes can't update the hash themselves, an important security constraint.
+    parentIDTemplate: 'spiffe://{{ .TrustDomain }}/spire/agent/tpm/{{ index .NodeMeta.Labels "node-restriction.kubernetes.io/tpm-pubhash" }}{{ index .NodeMeta.Labels "node-restriction.kubernetes.io/tpm-pubhash2" }}'
+    tpmDirect:
```
kfox1111, line 13:
```suggestion
  nodeAttestor:
    tpmDirect:
```

examples/tpm-direct/values.yaml
```diff
@@ -0,0 +1,15 @@
...
+    k8sPsat:
+      enabled: false
+    tpmDirect:
+      enabled: true
+
+spire-server:
+  controllerManager:
+    # K8s labels have a 63 character limit. TPM hashes are 64 chars. So you need to label the node with two labels with half of the tpm's hash each. The 'node-restriction.kubernetes.io/' prefix is so that the
+    # nodes can't update the hash themselves, an important security constraint.
+    parentIDTemplate: 'spiffe://{{ .TrustDomain }}/spire/agent/tpm/{{ index .NodeMeta.Labels "node-restriction.kubernetes.io/tpm-pubhash" }}{{ index .NodeMeta.Labels "node-restriction.kubernetes.io/tpm-pubhash2" }}'
```
marcofranssen, line 12: Should the chart check this `restriction.kubernetes.io` is part of the template as some sort of safetynet?

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -166,6 +166,37 @@ nodeAttestor:
   k8sPsat:
     ## @param nodeAttestor.k8sPsat.enabled Enable Psat k8s Node Attestor
     enabled: true
+  tpmDirect:
```
marcofranssen, line 169: Maybe just name this. Trying to understand what Direct means in this context.

```suggestion
  tpm:
```

marcofranssen (commented):
Small question and suggestion. Looks pretty neat.

charts/spire/charts/spire-agent/values.yaml
kfox1111: Marked as Direct to keep room for a second feature in the future. There is a second tpm mode that spire supports here: https://github.com/spiffe/spire/blob/main/doc/plugin_agent_nodeattestor_tpm_devid.md

Direct uses the tpm's private key hash directly for identification. This is the simplest way to use a tpm.

DeviceID requires an extra server and an out of band registration mechanism. But is the only way to attest some devices (Apple I think)

examples/tpm-direct/values.yaml
kfox1111: I wish we could, but it is a restriction on the labels being manually added to the node by the end user. Its not done in the chart, so not something we can check in the chart.

k8s itself does enforce the restriction.If you try and load it all in one string, it complains that the label is too long.

Once the tpm plugin pr's merge, I'm going to provide documentation at https://github.com/spiffe/spiffe.io/pull/290

charts/spire/charts/spire-agent/values.yaml
marcofranssen: ok, thanks for explaining, makes sense to keep room for extension. Is `direct` a uniformal way of referring to this tpm approach? If so I'm fine. Just wanter to pick your brain on the naming convention.

examples/tpm-direct/values.yaml
marcofranssen: Yes outside of the chart an enduser could do that indeed. But this value specifically we can control it right?

charts/spire/charts/spire-agent/values.yaml
kfox1111: I don't know of any standard naming around it. Also open to different names. Though naming things is hard.

examples/tpm-direct/values.yaml
kfox1111: Oh. I think I may have been misunderstanding what you've been saying...  I thought you were talking about the character limit, not the parentIDTemplate itself.

That would be good to do if we could figure out a way to do it. I can't think of a way off hand to be able to do that, outside of writing our own gotl parser?

faisal-memon: @marcofranssen Are you asking just to check if the string `restriction.kubernetes.io` is present?

charts/spire/charts/spire-agent/values.yaml
marcofranssen: Yup agree. I'm not very into tpm stuff so just asked my 101 questions here on the PR. I guess it is good to go. @faisal-memon feel free to approve and merge if you don't see any other hurdles.

examples/tpm-direct/values.yaml
marcofranssen: it was about the char limit indeed, but seeing Kevins response that is not possible prior to parsing the template I guess. So no worries, I'm fine merging like this. Just tried to see if we could failsafe the user.

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -41,13 +46,55 @@ spec:
...
+            - sh
+            - -c
+            - |
+              TPM=`/app/get_tpm_pubhash 2> /dev/null`
+              if [ $? -eq 0 ]; then
+                echo "TPM Fingerprint: $TPM"
+                (echo -n '<1>'; echo "TPM Fingerprint: $TPM") | tee /dev/kmsg
+              else
+                /app/get_tpm_pubhash
+              fi
```
faisal-memon, line 65: Can you add a comment here explaining what this snippet does?

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -41,13 +46,55 @@ spec:
...
+        {{- if .Values.nodeAttestor.tpmDirect.pubHash.enabled }}
+        - name: fingerprint-tpm
+          securityContext:
+            {{- $mainSecurityContext | toYaml | nindent 12 }}
+          image: {{ template "spire-lib.image" (dict "appVersion" $.Chart.AppVersion "image" .Values.nodeAttestor.tpmDirect.pubHash.image "global" .Values.global) }}
+          command:
+            - sh
+            - -c
+            - |
+              TPM=`/app/get_tpm_pubhash 2> /dev/null`
```
kfox1111, line 59:
```suggestion
              # Fetch the TPM's fingerprint from the TPM and output it on the Computers Console. This enables secure bootstrapping of the Node with the Server.
              TPM=`/app/get_tpm_pubhash 2> /dev/null`
```

faisal-memon (approved):

charts/spire/charts/spire-agent/values.yaml
faisal-memon: Naming looks ok to me. Keeping room for the second tpm plugin makes sense

PR #218, opened 2024-01-28
Support hybrid nodes with different attestors or config

kfox1111:
In some clusters, you may have a mix of nodes with differing attestor or config
needs. For example, some nodes have hardware TPMs and some without.
This patch enables configuring multiple daemonsets for the agent
that you can target to pools of nodes.


charts/spire/charts/spire-agent/templates/pre-upgrade-hook.yaml
```diff
@@ -0,0 +1,77 @@
+{{- if eq ((dig "installAndUpgradeHooks" "enabled" .Values.installAndUpgradeHook.enabled .Values.global) | toString) "true" }}
```
kfox1111, line 1: This is a one time hook that can be removed in 0.19.0

charts/spire/README.md
```diff
@@ -75,8 +75,9 @@ kubectl delete crds clusterfederatedtrustdomains.spire.spiffe.io clusterspiffeid
 
 We only support upgrading one major version at a time. Version skipping isn't supported.
 
-### 0.18.x
+### 0.18.X
```
faisal-memon, line 78:
```suggestion
### 0.19.X
```

faisal-memon (approved):

PR #223, opened 2024-01-30
Update README.md put cleanup seperately

cccsss01:


charts/spire/README.md
```diff
@@ -68,6 +68,14 @@ helm upgrade --install -n spire-mgmt spire spire --repo https://spiffe.github.io
 
 We only support upgrading one major version at a time. Version skipping isn't supported.
 
+### Clean up
```
kfox1111, line 71: Didn't notice before, but the cleanup section is nested here inside the middle of the upgrade notes section. It needs to be moved out to its own section.

cccsss01: Isn't this moved out with the 3 hashes?

charts/spire/README.md
```diff
@@ -64,6 +64,14 @@ helm upgrade --install -n spire-mgmt spire-crds spire-crds --repo https://spiffe
 helm upgrade --install -n spire-mgmt spire spire --repo https://spiffe.github.io/helm-charts-hardened/ -f your-values.yaml
 ```
 
+### Clean up
```
kfox1111, line 67:
```suggestion
## Clean up
```

kfox1111: Back to 2 so it is its own section at the same level as 'Install Instructions' and 'Upgrade notes' and then I think its good to go.

kfox1111 (approved):
LGTM. Thanks! :)

PR #227, opened 2024-01-31
Fix to apply securityContext recommendation to spire-agent init container

marcofranssen:
Attempt to add the securityContext back to this init container. However the macro function doesn't cater for being able to override the securityContext at spire-agent.waitForIt.securityContext.

@kfox1111 any thoughts on what would be required to fix this usecase for spire-agent init containers and probably others that we have missed, so the recommendation applies it on all container as opposed to just a selection?

### DIFF
--- a/charts/spire/charts/spire-agent/templates/daemonset.yaml
+++ b/charts/spire/charts/spire-agent/templates/daemonset.yaml
@@ -49,7 +49,7 @@ spec:
           resources:
             {{- toYaml .Values.waitForIt.resources | nindent 12 }}
           securityContext:
-            {{ toYaml .Values.securityContext | nindent 12 }}
+            {{- include "spire-lib.securitycontext" . | nindent 12 }}
         {{- if gt (int (dig "fsGroup" 0 $podSecurityContext)) 0 }}
         - name: fsgroupfix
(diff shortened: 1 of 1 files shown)


kfox1111 (commented):
Hmm... Will do some testing of this.

I initially avoided using the macro here as it has stronger restrictions then we ever have had on the agent pod. But I think your right that on the waitForIt container in particular, its probably ok to have it restricted tighter. It doesn't need to do anything fancy.

kfox1111:
Also, have a look at https://github.com/spiffe/helm-charts-hardened/pull/216. It has some containers that need permissions all the way up to 'privileged' in order to talk to the tpm. so may help inform the conversation as well.

faisal-memon:
@marcofranssen Do you want to cover other fixes here or just merge this and handle the other issues in a seperate pr?

marcofranssen:
we could move ahead with this fix only. Haven't detected other places for my own usecase that where not covered by the recommendation.

kfox1111:
> we could move ahead with this fix only. Haven't detected other places for my own usecase that where not covered by the recommendation.

kk. Well, the container involved in this pr was removed entirely in:
https://github.com/spiffe/helm-charts-hardened/pull/301

So, I think in that case we're done, and can just close out this pr?

Issue #224, opened 2024-01-30
Ensure that the agent admin socket can be enabled


edwbuck:
The Agent configuration supports an agent level setting agent_admin_socket which describes the Unix Domain Socket for Agent administration APIs.

Recent work on configurable log levels will place the post-log level adjustment commands on this socket.  It would be nice to have this setting be supported.

Support includes:

- having a pod-private volume to hold the socket
- ensuring that this volume is mounted in a directory higher than the workload API Unix Domain Socket (to satisfy the configuration file sanity checks in SPIRE)
- ensuring that the pod-private socket is exposed to the pod that holds the spire-agent executable, so the agent executable can be used as a client of the socket to adjust the already-running agent process.

kfox1111:
Does the spire-agent executable look for the admin socket in any particular place by default or does it look at the default config file and read it from there?

kfox1111:
Somewhat related to https://github.com/spiffe/helm-charts-hardened/issues/177 as well.

kfox1111:
Looking around, cilium is using the admin socket from their own pod for the delegated identity api. So we also need the ability to hostPath export it.

kfox1111:
https://github.com/spiffe/spire/pull/4856

kfox1111:
pr 4856 merged, so the default will be:
"/tmp/spire-agent/private/admin.sock"

Issue #177, opened 2024-01-10
spire-agent: cli commands do not function as expected


drewwells:
The default setup of this container does not align with the defaults assumed by the cli. As a result, most `kubectl exec` commands fail unless somebody has specific knowledge of the correct overrides. I think the best approach would be to place the config.yaml into a default location so other things like `socket-path` can be derived from loading the config file. Here's some examples of commands that do not work

I believe /run/spire/agent-socket/api.sock was in an example somewhere, but the actual socket name is `spire-agent.sock` as specified in the default config file
```
Defaulted container "spire-agent" out of: spire-agent, init (init)
unable to validate JWT SVID: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /run/spire/agent-sockets/api.sock: connect: no such file or directory"
command terminated with exit code 1
```

Or if you don’t specify a socketPath, you get an error like this
```
unable to validate JWT SVID: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial unix /tmp/spire-agent/public/api.sock: connect: no such file or directory"
```

Then if I try to find the socket filename on the container filesystem, I’m stuck on the environment's limited cli tools
```
error: Internal error occurred: error executing command in container: Internal error occurred: error executing command in container: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "7870b8cd1679b1a01cdc0e26edace55010d7b20fbc815ea9f3f7d9337c936ee1": OCI runtime exec failed: exec failed: unable to start container process: exec: "ls": executable file not found in $PATH: unknown
```

*A/C*
- spire-agent cli commands work without specifying the config or socket path locations

kfox1111:
There is a few different issues here.... Agreed that this could be made better in several ways.

There is no standard for what the socket name should be and last I looked, no way to specify it via environment variable. But we could probably wire something up to the default location it looks in so that the command works without passing the socket. This is problematic though if the user can update the socket name which they can. It probably needs https://github.com/spiffe/helm-charts-hardened/issues/176

We added /opt/spire/bin to the path already on the spire-server to make it easier to execute the cli via kubectl exec. We should do the same here.

The suggestion of setting the default entrypoint cant be implemented. kubectl exec does not honor container entrypoint by design. But the path thing should at least make it easier to find. so it can be :
```
kubectl -n spire-server exec daemonset/spire-agent -- spire-agent api validate 
```
rather then:
```
[...]

Issue #176, opened 2024-01-09
socket aliases


kfox1111:
Seems like istio expects the socket name to be 'socket' within the csi driver. But other things are expecting the socket name to be spire-agent.sock. We should test out if relative symlinks work, and if so, allow the user to specify a list of alias's for the socket name to be created in the socket share directory linked relatively to the 'spire-agent.sock'.

kfox1111:
So far, we have 3 differnet names I've found in common use:
* spire-agent.sock - chart default
* api.sock - spire-agent default
* socket - istio

We should probably support all 3 names out of the box.

PR #234, opened 2024-02-04
Add support for enabling the spire-agent admin socket

kfox1111:
Fixes: https://github.com/spiffe/helm-charts-hardened/issues/224


charts/spire/charts/spire-agent/values.yaml
```diff
@@ -257,3 +257,12 @@ experimental:
...
   featureFlags: []
+
+sockets:
+  ## @param sockets.hostBasePath Path on which the agent socket is made available when admin.expose is true
+  hostBasePath: /run/spire/agent/sockets
+  ## @param sockets.admin.enabled Enable the admin socket. Useful for admin tasks or the Delegated Identity API.
+  ## @param sockets.admin.expose Enable the admin socket to be visible on the host.
+  admin:
+    enabled: false
+    expose: false
```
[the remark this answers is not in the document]
kfox1111: Done

charts/spire/charts/spire-agent/README.md
```diff
@@ -109,3 +109,6 @@ A Helm chart to install the SPIRE agent.
...
 | `experimental.syncInterval`                       | Sync interval with SPIRE server with exponential backoff                                                                                                                                          | `5s`                                                                             |
 | `experimental.featureFlags`                       | List of developer feature flags                                                                                                                                                                   | `[]`                                                                             |
+| `sockets.hostBasePath`                            | Path on which the agent socket is made available when admin.expose is true                                                                                                                        | `/run/spire/agent/sockets`                                                       |
```
marcofranssen, line 112: Should this comment be referring to `/tmp/spire-agent/private/admin.sock`?

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -281,3 +281,12 @@ experimental:
   syncInterval: 5s
   ## @param experimental.featureFlags [array] List of developer feature flags
   featureFlags: []
+
+sockets:
+  ## @param sockets.hostBasePath Path on which the agent socket is made available when admin.expose is true
+  hostBasePath: /run/spire/agent/sockets
```
marcofranssen, line 287: See above.

charts/spire/charts/spire-agent/README.md
kfox1111: It needs to always be /tmp/spire-agent/private/admin.sock within the container so the cli can find it.

Outside the container, on the host, it should be configurable.

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -20,6 +20,9 @@
 {{- end }}
 {{- define "spire-agent.yaml-config" -}}
 agent:
+  {{- if .Values.sockets.admin.enabled }}
+  admin_socket_dir: /tmp/spire-agent/private/admin.sock
```
edwbuck, line 24: Private conveys the right security considerations, but not the reason one would use the socket.

Let's try to name things according to how they provide utility to the people using them, and not based on other characteristics that don't convey utility.  To clarify what that means, I don't go to a car dealership to buy "secure" I go to a car dealership to buy "transportation" that I could lock up (rendering it secure).  If I used this example in a naming pattern, "/tmp/spire-agent/transporation/car.sock" is preferred over "/tmp/spire-agent/secure/car.sock".

Please consider "/tmp/spire-agent/admin-api/admin.sock" or "/tmp/spire-agent/admin/admin.sock" or some other similar directory path instead of "/tmp/spire-agent/private/admin.sock"

charts/spire/charts/spire-agent/README.md
```diff
@@ -109,3 +109,6 @@ A Helm chart to install the SPIRE agent.
...
+| `sockets.admin.enabled`                           | Enable the admin socket. Useful for admin tasks or the Delegated Identity API.                                                                                                                    | `false`                                                                          |
+| `sockets.admin.mountOnHost`                       | Enable the admin socket to be visible on the host.                                                                                                                                                | `false`                                                                          |
```
edwbuck, line 114: This seems to control the placement of the socket on the host using hostPath semantics.  I'm not sure if we want to support hostPath mounting, as the only client that can support communicating to the socket is the one embedded in the `spire-agent` executable.

Please considering removing this option / configuration approach.

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -170,6 +177,15 @@ spec:
         - name: spire-config
           configMap:
             name: {{ include "spire-agent.fullname" . }}
+        {{- if .Values.sockets.admin.mountOnHost }}
```
edwbuck, line 180: The client for the spire-agent UDS is the executable `spire-agent`.  I don't think we should expose the socket using `hostPath` as the only deployed client will exist within the spire-agent Pod.

charts/spire/charts/spire-agent/README.md
edwbuck: I don't think there is a valid use case for putting the socket outside of the Pod.  The client that works with the socket is the `spire-agent` executable, called in a non-"run" context.

charts/spire/charts/spire-agent/values.yaml
edwbuck, line 287: I don't think there is a valid use case for putting the admin socket on the host.

edwbuck (commented):
I don't see any checks to ensure that the agent admin socket is not in the same directory (or a sub directory) of the agent socket.

Considering that when someone configures this with the agent admin socket in the same directory or below the agent socket directory the agent won't launch, I think we should handle this misconfiguration in the helm template (rather than let them publish the wrong values and watch their agents enter a crash loop).

charts/spire/charts/spire-agent/README.md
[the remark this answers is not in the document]
kfox1111: This set of values is leaving room in the future to fix the agent api socketPath option to also reuse sockets.hostBasePath for easier configuration. I didn't do that refactor here though as its a lot more changes then adding a new socket.

charts/spire/charts/spire-agent/README.md
kfox1111: The admin socket is also used for the "delegated identity api". At least one major software is needing to use it. Cilium. hostPath is needed to support their needs.

charts/spire/charts/spire-agent/templates/daemonset.yaml
kfox1111: Addressed above

charts/spire/charts/spire-agent/values.yaml
kfox1111: Addressed above

charts/spire/charts/spire-agent/README.md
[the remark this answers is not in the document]
edwbuck: Perhaps the future fix can include the renaming across all elements?  The commit as-is will have a confusing setup where some elements are configured via `socket` and others `sockets`.  I'd feel more comfortable putting the "future fix" stuff in a different commit that solely addresses the goals of the future fix, or expanding the scope to fully encompass the future fix such that all of it is delivered in this PR.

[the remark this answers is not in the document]
faisal-memon: What do you think of moving this config under sockets.admin so its clear it applies only to the admin socket? Also need to change admin.expose -> admin.mountOnHost

charts/spire/charts/spire-agent/templates/configmap.yaml
edwbuck: #176 
Istio should be using the workload api socket, aka the agent socket, so alias / naming of sockets for the purposes of istio compatibility seem out of place in considering the agent admin socket.

#177 
As far as I'm concerned, if one is attempting to run the spire-agent executable, they can pass on the command line the configuration file that will tell the executable where the config file exists, the same pattern we use in the Agent's primary Pod.  That config file will work even when the executable is being called as a client for a previously launched Agent service.

Neither of these items seem to have an impact on what we call the directory.

charts/spire/charts/spire-agent/README.md
edwbuck: Do we support Cilium at this time?  Why are we complicating this PR with items outside the scope of adding the Agent Admin Socket?

amoore877: > The admin socket is also used for the "delegated identity api". At least one major software is needing to use it. Cilium. hostPath is needed to support their needs.

fwiw, my org also makes use of a host-mounted being available for delegated identity. Maybe that's something expected in flex?

> Why are we complicating this PR with items outside the scope of adding the Agent Admin Socket?

I'm generally all for keeping a set of changes as simple as possible, as it has benefits to commit history and reviewer burden, and meshes well with keeping issues as bite-sized as possible for easy wins.
It would be helpful to clarify desired results of #224, and how they seemed to expand due to investigations posted in that issue.
It can also be helpful in online communication where we lack the ability to see tone and expressions to post reasoning/opinions declaratively as much as possible ("I don't think we need foo here, because it is outdated and replaced with bar"), which can then be argued for or against, rather than interrogatively ("Why are we using foo?"), which is more difficult to continue discussion with.

faisal-memon:
> I don't see any checks to ensure that the agent admin socket is not in the same directory (or a sub directory) of the agent socket.

@kfox1111 I think this is a good one to include. Do you have the bandwidth to address? If not I can push a commit that does this check.

charts/spire/charts/spire-agent/README.md
faisal-memon: @edwbuck Thanks for highlighting this. I talked to @kfox1111 offline and also questioned why this was needed. Upon hearing the cillium use case, it being disabled by default, and the work already being done I determined it was best to leave it in place. Definitely hear the concerns about scope creep, a good reminder to more clearly define requests in the future.

kfox1111: @amoore877 Totally understand about small pr's. There are certainly advantages. But we've also had difficulty with pr merge conflicts. So its a difficult balancing act between putting too little in a pr, causing too many pr's, and constantly dealing with merge conflicts vs having too big pr's to easily review. This size pr has been generally a reasonable compromise in the past.

kfox1111:
> > I don't see any checks to ensure that the agent admin socket is not in the same directory (or a sub directory) of the agent socket.
> 
> @kfox1111 I think this is a good one to include. Do you have the bandwidth to address? If not I can push a commit that does this check.

I have added this test.

charts/spire/charts/spire-agent/README.md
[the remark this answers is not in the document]
edwbuck: I have mixed feelings.  If we get into the habit of accepting work in the wrong contexts, I worry that perhaps we will find it hard to roll back to denying work that is "directionally correct" in issues that have nothing to do with the item under review.

Under the worst cases, this means that we can't review with a scope in mind, because only some of the work will be present in the issue that eventually has the stage when the final bits of work are completed.  In this project and others, we've found bugs in the "pre submitted" work that wasn't caught in review under the final work, because the final work didn't have the controls of indicating all the changed files tied to the issue.

And the "ends justify the means" handling of it here rubs me wrong.

That said, I did say I'd align to your wishes, and in the end, if it's how you work, I'm not going to alter it.

Approved

charts/spire/charts/spire-agent/templates/configmap.yaml
edwbuck: I'm fine with the name as it stands then.  This was a call to improve upon the existing path, but there's value in matching the path to documentation elsewhere.

charts/spire/charts/spire-agent/README.md
edwbuck: @kfox1111 we would have less issue with merge conflicts if the merges were targeted to only the items being changed, as they would be smaller in footprint and less likely to overlap.

What you're offering is too big of a PR with too many PRs and you're getting items from one PR mixed into others.  This is independent of how many PRs are available.  These kinds of "mixing issues" make it difficult to review the PR which is why your PRs take longer to review and get more pushback.

If each PR was small, sweet, and to the point, then reviews would go faster, fewer PRs would remain open, and you'd have less reason to see many PRs with merge conflicts.

charts/spire/charts/spire-agent/templates/daemonset.yaml
edwbuck: I'd rather see this added with Cilium's PR, so we can review it all in one context that has something to do with Cilium.  When we review Cilium supporting changes, because this will already be merged into the baseline, we likely will miss any issues that tie back to this change.

edwbuck (approved):
approved

faisal-memon (approved):

Issue #217, opened 2024-01-26
what to do with examples/production


kfox1111:
examples/production/ is not as useful as it was in the past. The vast majority of things it did are now in the chart now via recommendations. The documentation facet of it is better handled via chart/README.md and the spiffe.io website. And the test functionality is a little counter to it being a good example.

So, we should figure out what to do with it.

Currently thinking of moving it to tests/integration/production

marcofranssen:
Some things to handle when moving:

- [ ] Update workflow regarding examples test
- [ ] Update branch protections to require another mandatory job (remove examples/production with new job)

PR #237, opened 2024-02-06
Change production example to be an integration test

kfox1111:
The documentation is a better example then the test now. Separate the two.

Fixes https://github.com/spiffe/helm-charts-hardened/issues/217

Related to: https://github.com/spiffe/helm-charts-hardened/pull/206


marcofranssen (commented):
Maybe we should decouple the examples from the integration `values.yaml`. For the purpose of an example it would be more transparant to find that values.yaml in the same folder. Maybe we should decouple that part and just go with the default values for an example and just only define the values required for that specific example.

kfox1111:
> Maybe we should decouple the examples from the integration `values.yaml`. For the purpose of an example it would be more transparant to find that values.yaml in the same folder. Maybe we should decouple that part and just go with the default values for an example and just only define the values required for that specific example.

Hmm... What if we moved the run-tests.sh's of all of these tests into the tests/integration subdir instead of examples then link back to the values in examples?  Then the users would be less likely to look at the run-tests.sh and get confused by it while still allowing us to test production configurations of the example values?

kfox1111:
@marcofranssen I tried to take a stab at incorporating your feedback. Please rereview when you get a chance.

.github/tests/common.sh
```diff
@@ -70,3 +70,23 @@ $(helm ls -A | sed 's/\t/ | /g' | sed 's/^/| /' | sed 's/$/ |/' | sed '/^| NAME.
...
+global:
+  spire:
+    recommendations:
+      enabled: true
+    clusterName: production
+    trustDomain: production.other
+    caSubject:
+      country: US
+      organization: Production
+      commonName: production.other
```
faisal-memon, line 86: Why have the values like this rather than in a file?

faisal-memon (approved):

PR #242, opened 2024-02-08
Complete Server K8S PSAT support

kfox1111:
Add all the SPIRE Server supported options for the K8S PSAT attestor. This retains the ease of use for configuring local cluster support while adding the ability to configure multiple/external clusters as well. Kubeconfig support is added in its own config block as it will be used/shared with spire-controller-manager support in the future.


charts/spire/charts/spire-server/values.yaml
```diff
@@ -683,6 +683,29 @@ nodeAttestor:
...
+    audience: ["spire-server"]
+    ## @param nodeAttestor.k8sPsat.allowedNodeLabelKeys [array] Node label keys considered for selectors
+    allowedNodeLabelKeys: []
+    ## @param nodeAttestor.k8sPsat.allowedPodLabelKeys [array] Pod label keys considered for selectors
+    allowedPodLabelKeys: []
+  externalK8sPsat:
+    ## @param nodeAttestor.externalK8sPsat.enabled Enable Psat k8s nodeattestor for external Kubernetes clusters
+    enabled: false
+    defaults:
+      ## @param nodeAttestor.externalK8sPsat.defaults.serviceAccountAllowList [array] Allowed service accounts for Psat nodeattestor
```
faisal-memon, line 696:
```suggestion
      ## @param nodeAttestor.externalK8sPsat.defaults.serviceAccountAllowList [array] Allowed service accounts for PSAT node attestor
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -685,6 +685,29 @@ nodeAttestor:
...
     ## @param nodeAttestor.k8sPsat.serviceAccountAllowList [array] Allowed service accounts for Psat nodeattestor
     serviceAccountAllowList: []
+    ## @param nodeAttestor.k8sPsat.audience [array] Audience for token validation. If it is set to an empty array ([]), Kubernetes API server audience is used
+    audience: ["spire-server"]
+    ## @param nodeAttestor.k8sPsat.allowedNodeLabelKeys [array] Node label keys considered for selectors
+    allowedNodeLabelKeys: []
+    ## @param nodeAttestor.k8sPsat.allowedPodLabelKeys [array] Pod label keys considered for selectors
+    allowedPodLabelKeys: []
+  externalK8sPsat:
+    ## @param nodeAttestor.externalK8sPsat.enabled Enable Psat k8s nodeattestor for external Kubernetes clusters
```
faisal-memon, line 695:
```suggestion
    ## @param nodeAttestor.externalK8sPsat.enabled Enable PSAT k8s nodeattestor for external Kubernetes clusters
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -685,6 +685,29 @@ nodeAttestor:
...
+    enabled: false
+    defaults:
```
faisal-memon, line 697: Why the extra `defaults` level?

kfox1111: Also, the defaults for the internal cluster can be different from external ones. For example, the serviceAccountAllowList will always need to be different in the external case.

marcofranssen: Does the doc tool allow to define a doc at the defaults level? Maybe we could add the explanation there to make it clear. Not sure if that will actually render in the md file though.

marcofranssen (approved):
Didn't test, but I see automated tests that cover the feature.

🚀  LGTM

charts/spire/charts/spire-server/values.yaml
kfox1111: Yeah, I don't think we can get it into the md file nicely. Its a case we can document really well in the spiffe.io docs though since we have more control over it, so maybe thats enough?'

charts/spire/charts/spire-server/values.yaml
```diff
@@ -697,6 +697,29 @@ nodeAttestor:
     enabled: true
     ## @param nodeAttestor.k8sPsat.serviceAccountAllowList [array] Allowed service accounts for Psat nodeattestor
     serviceAccountAllowList: []
+    ## @param nodeAttestor.k8sPsat.audience [array] Audience for token validation. If it is set to an empty array ([]), Kubernetes API server audience is used
```
faisal-memon, line 700:
```suggestion
    ## @param nodeAttestor.k8sPsat.audience [array] Audience for token validation. If set to [] (empty array), Kubernetes API server audience is used
```

charts/spire/charts/spire-server/values.yaml
faisal-memon: Is this because its an array and we cant set defaults for arrays?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -697,6 +697,29 @@ nodeAttestor:
...
+      audience: ["spire-server"]
+      ## @param nodeAttestor.externalK8sPsat.defaults.allowedNodeLabelKeys [array] Node label keys considered for selectors
+      allowedNodeLabelKeys: []
+      ## @param nodeAttestor.externalK8sPsat.defaults.allowedPodLabelKeys [array] Pod label keys considered for selectors
+      allowedPodLabelKeys: []
+    ## @param nodeAttestor.externalK8sPsat.clusters [object] A dictionary of clusters to add with optional overrides. If empty, all clusters defined in kubeconfigs will be used.
+    clusters: {}
+    # clustera:
+    #   serviceAccountAllowList: ["other-ns:other-agent"]
+    # clusterb: {}
```
faisal-memon, line 722: I think its better to link this to the `kubeconfigs` section with a key rather than requiring they have the same name. Similar to how `serviceAccountName` in k8s works.

charts/spire/charts/spire-server/values.yaml
```diff
@@ -884,3 +907,12 @@ tests:
       repository: chainguard/bash
       pullPolicy: IfNotPresent
       tag: latest@sha256:5c232c444e5c3bc509fdd3794534e9ff5438e9f80268218202ce6ce8c6d8e8f7
+
+## @param kubeconfigs [object] Manage additional kubeconfig files to talk to external Kubernetes clusters
+kubeconfigs: {}
```
faisal-memon, line 912: Should it be`kubeConfig`?

charts/spire/charts/spire-server/values.yaml
kfox1111: so, like:
```
clusters:
  clustera:
    kubeConfig: foo
    
kubeConfigs:
  foo:
    kubeConfig: abcdefg
```

?

faisal-memon: Yea, or `kubeConfigName` to match the `serviceAccountName` pattern.

charts/spire/charts/spire-server/values.yaml
kfox1111: Done

charts/spire/charts/spire-server/values.yaml
kfox1111: Done

faisal-memon (approved):

Issue #250, opened 2024-02-15
It seems the tools.kubectl.image is not globally overrideable


marcofranssen:
I have been trying to globally override the `globals.tools.kubectl.image.tag`.

Either we need to improve the documentation or we have a bug in that helper function or it usage.

kfox1111:
Yeah, being able to override it everywhere at once makes a lot of sense. Will make a pr for this.

PR #251, opened 2024-02-18
Add global override for kubectl tag

kfox1111:
If you need to update kubectls tag, you need to do it in a lot of places currently. This allows setting it in one place.

Fixes: https://github.com/spiffe/helm-charts-hardened/issues/250


marcofranssen (approved):
LGTM 🚀

PR #260, opened 2024-02-22
Fix OpenShift Federation Ingress bug

maia-iyer:
Working on demoing SPIRE Federation on OpenShift - Helm chart currently fails on deployment with following helm values on Openshift:

```
spire-server:
  ingress:
    enabled: true
  federation:
    enabled: true
    ingress:
      enabled: true
```

With error:

```
Error: template: spire/charts/spire-server/templates/federation-ingress.yaml:28:14: executing "spire/charts/spire-server/templates/federation-ingress.yaml" at <"">: undefined variable: $path
```

### DIFF
--- a/charts/spire/charts/spire-server/templates/federation-ingress.yaml
+++ b/charts/spire/charts/spire-server/templates/federation-ingress.yaml
@@ -14,7 +14,6 @@
 {{-   $_ := set $federationIngress "tlsSecret" .Values.federation.tls.externalSecret.secretName }}
 {{- end }}
 {{- $ingressControllerType := include "spire-lib.ingress-controller-type" (dict "global" .Values.global "ingress" .Values.federation.ingress) }}
-{{- $tlsSection := true }}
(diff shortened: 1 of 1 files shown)


charts/spire/charts/spire-server/templates/federation-ingress.yaml
```diff
@@ -1,6 +1,9 @@
 {{- if .Values.federation.enabled }}
 {{- if .Values.federation.ingress.enabled -}}
 {{- $svcName := include "spire-server.fullname" . }}
+{{- $path := "/"}}
+{{- $pathType := "Prefix" }}
+{{- $tlsSection := true }}
```
kfox1111, line 6: Yeah, good catch.

I think there is one more issue to resolve and I think it looks good to me. I can't do an inline suggestion unfortunately.

Down a few lines (14 / 17) is a duplicate:
```
{{- $tlsSection := true }}
```
That should probably be removed.

kfox1111 (approved):
lgtm. Thanks! :)

PR #266, opened 2024-02-28
Fix issue with tests not running  on automated PRs

### LINKED ISSUE #10: Automated version update PRs not running tests
When creating automated PRs from the cron jobs that checks versions, the tests arent being run automatically. We have to push a commit to the branch to force the tests to run.

Suggestion is to use a GitHub App .

faisal-memon:
Use a github app to generate token for the version checker to use when creating PRs. Draft until SSC adds the github app to this repo. Currently using a github app in personal account. Will request official app in spire org after validating this works ok.

fixes #10

### DIFF
--- a/.github/workflows/check-versions.yaml
+++ b/.github/workflows/check-versions.yaml
@@ -60,7 +60,7 @@ jobs:
           git diff
 
       - name: Generate Token
-        uses: tibdex/github-app-token@v1
+        uses: tibdex/github-app-token@v2.1.0
         id: generate-token
         with:
           app_id: ${{ vars.APP_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/check-versions.yaml
```diff
@@ -59,11 +59,18 @@ jobs:
           ./.github/scripts/update-tags.sh
           git diff
 
+      - name: Generate Token
+        uses: tibdex/github-app-token@v1
```
marcofranssen, line 63: Can we lock this on the patch release so we are in control when we update it.

kfox1111 (approved):
Looks good to me. Thanks!

PR #271, opened 2024-03-01
add pod labels to spire server values

mcrors:
The reason for this change is to be able to integrate spire with Azure Workload Idenity. To enable this, a label of `azure.workload.identity/use: "true"` needs to be added to the pods metadata.

https://learn.microsoft.com/en-us/azure/aks/learn/tutorial-kubernetes-workload-identity#deploy-the-workload


charts/spire/charts/spire-server/templates/statefulset.yaml
```diff
@@ -43,6 +43,9 @@ spec:
         component: server
         release: {{ .Release.Name }}
         release-namespace: {{ .Release.Namespace }}
+        {{ with .Values.podLabels }}
```
marcofranssen, line 46:
```suggestion
        {{- with .Values.podLabels }}
```

marcofranssen (changes requested):
To remove the whiteline from the rendered yaml, please add the suggestion please.

charts/spire/charts/spire-server/templates/statefulset.yaml
mcrors: Cheers.

kfox1111 (approved):
LGTM. Thanks! :)

PR #273, opened 2024-03-04
Add pod labels to the SPIRE agent

mcrors:
Follow on from https://github.com/spiffe/helm-charts-hardened/pull/271. This will allow the spire agent to also integrate with azure workload identity.


kfox1111 (approved):
LGTM. Thanks! :)

PR #289, opened 2024-03-18
Add spire-lib chart

kfox1111:
Make spire-lib bits into its own library chart.


kfox1111:
Squashed.

charts/spire/templates/NOTES.txt
```diff
@@ -1,35 +1 @@
...
-{{-     if (index .Values "spire-server").controllerManager.watchClassless }}
-
-Spire CR's will be handled if no className is specified or if className is set to "{{ $className }}"
-{{-     else }}
-
-Spire CR's will be handled only if className is set to "{{ $className }}"
-{{-     end }}
-{{-   end }}
-{{- end }}
+{{ include "spire-lib.notes" . }}
```
marcofranssen, line 1: Wouldn't each chart define its own Notes specific to that chart?

E.g. once we break out the spire-agent, and other charts, each of those charts would have their own specific notes, then this reusable Notes thing seems a bit overkill as you would have to do all kind of if statements.

charts/spire/charts/spire-lib/values.yaml
marcofranssen, line 1: Can't we remove this dummy file?

charts/spire/charts/spire-lib/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.1.0"
```
marcofranssen, line 24:
```suggestion
appVersion: ""
```

the appVersion can be left empty. As the lib chart probably will not be dependant on a specific spire version and isn't installing an app I suggest to leave it empty.

marcofranssen (commented):
Some inline feedback.

charts/spire/templates/NOTES.txt
kfox1111: Yeah. I thought about that. Charts that look very similar though could use the same notes most likely. Those that deviate too much from the base chart could do their own notes and ignore the macro entirely though?

charts/spire/charts/spire-lib/values.yaml
kfox1111: No. The doc tool flips out. :/

charts/spire/templates/NOTES.txt
kfox1111: I reverted the notes bit.

faisal-memon: @marcofranssen Any further comments on this pr?

PR #301, opened 2024-03-27
Update to SPIRE 1.9.2, use retry_bootstrap, remove wait for it

faisal-memon:
Uses the new `retry_bootstrap` feature in 1.9.2 so we don't need wait for it in agent anymore.


charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -115,17 +115,6 @@ spec:
               mountPath: /tpm
           imagePullPolicy: {{ .Values.nodeAttestor.tpmDirect.plugin.image.pullPolicy }}
         {{- end }}
-        - name: init
-          # This is a small image with wait-for-it, choose whatever image
-          # you prefer that waits for a service to be up. This image is built
-          # from https://github.com/vishnubob/wait-for-it
-          image: {{ template "spire-lib.image" (dict "image" .Values.waitForIt.image "global" .Values.global) }}
```
kfox1111, line 122: The pr is looking pretty good.

I think its leaving behind a 'waitForIt' section in values.yaml. Could you remove that too please?

faisal-memon: Removed

kfox1111: Oh. Found one more reference:
https://github.com/spiffe/helm-charts-hardened/blob/main/.github/tests/images.json#L10-L14

faisal-memon: Thanks, fixed

kfox1111 (approved):
LGTM. Thanks! :)

PR #302, opened 2024-03-27
Add an easy to use identity for child servers

kfox1111:


kfox1111:
Discussed refactoring this to use type: raw and specify the selectors directly.

PR #303, opened 2024-03-27
Add support for externalServer=true

kfox1111:
This allows the chart to be used to manage identities, the bundle, permissions, etc without actually deploying the server. This is useful when integrating with an external server.


charts/spire/values.yaml
```diff
@@ -112,7 +112,7 @@ spire-server:
   enabled: true
   ## @param spire-server.nameOverride Overrides the name of Spire server pods
   nameOverride: server
-  ## @param spire-server.kind Run spire server as deployment/statefulset. This feature is experimental.
+  ## @param spire-server.kind Define SPIRE server kind. Can be one of [statefulset, deployment, none]. Setting to deployment is experimental.
```
faisal-memon, line 115:
```suggestion
  ## @param spire-server.kind Define SPIRE server kind. Can be one of [statefulset, deployment, none]. Setting to deployment is experimental. None will deploy everything except the server itself.
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -23,6 +23,9 @@ image:
 ## @param kind Define SPIRE server deployment type. Can be statefulset/deployment. Defaults to statefulset if not set. This feature is experimental.
 kind: statefulset
 
+## @param externalServer Set to true to setup the bundle configmap, rbac rules, and identity documents but doesn't deploy the server locally. Useful for external servers>
```
faisal-memon, line 26:
```suggestion
## @param externalServer Deploy only the bundle ConfigMap, RBAC rules, and identity documents but not the server. Use in a nested setup where the server is external.
```

faisal-memon (approved):

PR #305, opened 2024-03-27
nameOverride autoconfig support

kfox1111:
When setting up nested charts with nameOverride, you need to be able to inform other charts about the override.


charts/spire/charts/spire-server/values.yaml
```diff
@@ -390,6 +390,8 @@ upstreamAuthority:
     upstreamDriver: ""
     ## @extra upstreamAuthority.spire.server Server details for the Spire instance use as upstream CA
     server:
+      ## @param upstreamAuthority.spire.server.nameOverride Override the name for upstream Spire server
```
faisal-memon, line 393: Can we add a note on what the default will be if not overidden?

kfox1111: It might make the rendered doc string rather long.

Its mostly used if your assembling a chart out of the subcharts and set nameOverride on the root spire server. then it should be set to the same nameOverride here too to have the default autocalculated value work.

Its used to align the names here:
* https://github.com/spiffe/helm-charts-hardened/blob/nested-chart/charts/spire-nested/values.yaml#L214
* https://github.com/spiffe/helm-charts-hardened/blob/nested-chart/charts/spire-nested/values.yaml#L123

For end users, either the default will just work, or they will set server.address instead.

Maybe we add a small bit of text to the doc string stating it should only be changed when building your own nested chart?

faisal-memon: Thats fine too

kfox1111: kk. updated

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -109,11 +109,13 @@ upstream: false
 ## @param server.address Address for Spire server
 ## @param server.port Port number for Spire server
 ## @param server.namespaceOverride Override the namespace for Spire server
+## @param server.nameOverride Override the name for Spire server. Should only be changed when building your own nested chart
```
faisal-memon, line 112:
```suggestion
## @param server.nameOverride Override the name for Spire server. Should only be changed when building your own nested chart to ensure names align.
```

charts/spire/charts/spire-server/values.yaml
faisal-memon, line 393:
```suggestion
      ## @param upstreamAuthority.spire.server.nameOverride Override the name for upstream Spire server. Should only be changed when building your own nested chart to ensure names align.
```

faisal-memon (approved):

PR #307, opened 2024-03-28
Add global for storageClass

kfox1111:
When you have multiple spire servers, they often all need to be configured to use the same storage class. Let them all be set to the same value all at once.


charts/spire/values.yaml
```diff
@@ -29,6 +29,10 @@ global:
       organization: ""
       commonName: ""
 
+    persistence:
+      ## @param global.spire.persistence.storageClass What storage class to use for persistence
+      storageClass: null
```
faisal-memon, line 34: is it possible to put something other than null here? Like `default` or something?

kfox1111: I made it consistent with what is in tree already:
https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-server/values.yaml#L150

We could change it to a blank string safely I think, if you think that is better.

Any other value actually has meaning to Kubernetes. An undefined storageClass in k8s means to use whatever storage class name is set to be the default one by the cluster-admin (see https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/#changing-the-default-storageclass annotations for details), so we should not set it to the string 'default' as that will likely break on most clusters.

faisal-memon: i see. Since its already there I think its fine. Though I think empty string is better.

faisal-memon (approved):

PR #311, opened 2024-04-02
Add config for experimental events based cache

faisal-memon:
Experimental config in SPIRE to use events to hydrate SPIREs in memory cache rather than polling the the database.


charts/spire/charts/spire-server/values.yaml
```diff
@@ -909,6 +909,10 @@ experimental:
   enabled: false
   ## @param experimental.cacheReloadInterval The amount of time between two reloads of the in-memory entry cache.
   cacheReloadInterval: 5s
+  ## @param experimental.eventsBasedCache Use events to update the cache with what's changed since the last update.
+  eventsBasedCache: false
+  ## @param experimental.pruneEventsOlderThan How old an event can be before being deleted. Used with events based cache.
+  pruneEventsOlderThan: 12h
```
kfox1111, line 915: https://github.com/spiffe/spire/blob/main/doc/spire_server.md lists the default as 5s.

kfox1111 (commented):
Looks good to me except for maybe a deviation from the default.

charts/spire/charts/spire-server/values.yaml
faisal-memon: the cache reload default is 5s. Pruning events is 12 hours.

kfox1111: Doh. Your right. Sorry about that.

kfox1111 (approved):
LGTM. Thanks! :)

PR #316, opened 2024-04-05
Upgrade to spire-controller-manager 0.5.0

kfox1111:


kfox1111:
Draft until 0.4.4 is actually released

charts/spire/charts/spire-server/values.yaml
```diff
@@ -465,6 +465,9 @@ controllerManager:
   ## @param controllerManager.watchClassless specify to process custom resources without class name specified. Useful to slowly migrate to class names from classless installs. Do not have two installs on the same k8s cluster both set to true.
   watchClassless: false
 
+  ## @param controllerManager.entryIDPrefixCleanup consult the spiffe.io docs about this option before changing. Its unlikely you will need to ever change it.
```
faisal-memon, line 468: We should put the description of what it does along with the warning.

PR #322, opened 2024-04-08
Add resource limits for upgrade and delete hook batch jobs

anhpatel:


kfox1111:
+1 to a global setting for these. Switching it to only be global though may break some existing users.

Maybe use the sprig mergeOverwrite function to support both? (https://masterminds.github.io/sprig/dicts.html)

anhpatel:
@kfox1111 These global resource limits are only being set for all batch related containers. Which other case of resource limits do you think this will conflict with?

kfox1111:
Oh. I guess I saw a commit wiz by (https://github.com/spiffe/helm-charts-hardened/pull/322/commits/d099e572be8de5548cf276b34e1a9e26925bb61c) and thought it was changing existing code. Never mind my previous comment. Sorry for the confusion.

charts/spire/charts/spiffe-oidc-discovery-provider/templates/pre-delete-hook.yaml
```diff
@@ -72,4 +72,6 @@ spec:
           - deployment
           - {{ include "spiffe-oidc-discovery-provider.fullname" . }}
           - --wait
+        resources:
+          {{- toYaml .Values.global.deleteHooks.resources | nindent 10 }}
```
kfox1111, line 76: How about something like the following here and elsewhere:
```suggestion
        {{ with (dig "deleteHooks" "resources" (dict) .Values.global) }}
        resources:
          {{- toYaml . | nindent 10 }}
        {{ end }}
```

It would work when globals are not defined in the case where we are talking about having the option to install the individual charts standalone.

edwbuck: Each evaulation of a Helm value will return nil (nothing) provided it is referenced by a valid name.  Additionally, (nil).name yeilds (nil).  For this reason I would suggest the following approach to replace the often-confusing "dig"

Also I think "with (nil)" evaluates to no visitation of the loop body, please verify this.

{{ with ((.Values.global).deleteHooks).resources }} can replace both the approaches, and reads much nicer.

edwbuck (changes requested):
Better formatting for avoiding dig / missing tree elements suggested.  Apply liberally with testing to ensure the new approach doesn't break anything.

kfox1111:
We can't write to this repo, and we haven't heard back from the original author in a long time. Continuing development in https://github.com/spiffe/helm-charts-hardened/pull/366 instead since we can write to it there.

PR #344, opened 2024-05-02
Fix Tornjak logsDir for Openshift

mrsabath:
NPM used by Tornjak Frontend running on Openshift requires write access


kfox1111 (commented):
One minor thing and then I think its good to go

kfox1111 (approved):
LGTM. Thanks! :)

PR #347, opened 2024-05-03
Fix federation certificate name when upstream enabled

kfox1111:
When both federation certificates and upstream authority both use cert-manager, there is a naming conflict.


drewwells (approved):
100%, federated https_web is blocked on this

faisal-memon (approved):

PR #351, opened 2024-05-07
Add nested support via tags to main chart

kfox1111:
This explores using one chart for nested rather then a separate chart, using chart tags.

### DIFF
--- a/charts/spire/values.yaml
+++ b/charts/spire/values.yaml
@@ -144,9 +144,14 @@ spire-agent:
 # spiffe-csi-driver:
   # enabled: true
 
-# Used with tags [simple, nestedRoot, nestedChildFull, nestedChildSecurity]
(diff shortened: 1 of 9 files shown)


charts/spire/values.yaml
```diff
@@ -1,6 +1,16 @@
 # Default configuration for Spire chart
 # SPDX-License-Identifier: APACHE-2.0
 
+tags:
+  ## @param tags.simple Set the chart architecture to be a simple server/agent setup
```
kfox1111, line 5:
```suggestion
## @param tags.simple Set the chart architecture to be a single server and agent setup```

kfox1111:
closing in favor of https://github.com/spiffe/helm-charts-hardened/pull/294

kfox1111:
Closing in favor of https://github.com/spiffe/helm-charts-hardened/pull/294

PR #378, opened 2024-06-06
Apply wildcard for ignoreNamespaces in Controller Manager

mrsabath:
Also fix some spelling in the values.yaml file


kfox1111:
Looks like the docs script needs to be run still.

marcofranssen (approved):

edwbuck:
I don't think this was tested.  As I mentioned in #363 I could only find support for regex patterns in the Ignore Namespaces.

> Sorry, I realize that I did not answer one of your questions directly. Yes, "ignoreNamespace" does support wildcards, but only in the form of Regex patterns.

https://github.com/spiffe/spire-controller-manager/blob/00d745e6528190d04ce40d5a2543582dbf5103de/cmd/main.go#L61 should make my reasoning clear on why Regex is the only supported pattern engine.

kfox1111:
> I don't think this was tested.

It passed all tests before, and was only adding addition entries to a list that already was used for a long time.

That being said, switching to regexes is better.

edwbuck (approved):
@mrsabath Thanks for going the extra mile to also fix all of the spelling mistakes.

kfox1111 (approved):
LGTM. Thanks! :)

mrsabath:
Thank you @kfox1111 for your suggestions. I tested it earlier and it worked but with your changes it is even better!

PR #381, opened 2024-06-12
Update _spire-system-namespace.yaml

tuxotron:
system namespace was using the server labels


kfox1111 (approved):
Good catch.

LGTM. Thanks! :)

PR #384, opened 2024-06-17
Assign default, working rancher/kubectl version

mrsabath:
Currently the default rancher version does not match the existing image tags and that fails the install.
Failure on spire-server-post-install pod:
 ```
 Warning  Failed          50s (x3 over 91s)  kubelet            Failed to pull image "docker.io/rancher/kubectl:v1.28.10": reading manifest v1.28.10 in docker.io/rancher/kubectl: manifest unknown
  Warning  Failed          50s (x3 over 91s)  kubelet            Error: ErrImagePull
```


kfox1111 (changes requested):
-1

picking a very old kubectl that likely has cve's is not a good plan IMO. You can do it at runtime yourself and not hit any cve's you don't already have. But not something we should recommend to everyone to do out of the box.

PR #387, opened 2024-06-20
Update keycloak-config-cli using SPIRE example to Keycloak >= 24.0.0

moritzschmitz-oviva:
Keycloak >= 24.0.0 deprecated the use of password-protected truststores. This makes the integration with SPIRE quite a bit easier. This PR should update the example to reflect the changes.

### DIFF
--- a/examples/keycloak-config-cli-using-spire/README.md
+++ b/examples/keycloak-config-cli-using-spire/README.md
@@ -1,5 +1,10 @@
 # keycloak-config-cli using spire
 
+> [!WARNING]
+> This example uses
+> the [`SidecarContainers`](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#enabling-sidecar-containers)
+> feature. This is only enabled by default in Kubernetes 1.29+.
+
 This example shows how to leverage SPIRE in establishing an mTLS connection
 between [Keycloak (>= 24.0.0)](https://www.keycloak.org/) and [keycloak-config-cli](https://github.com/adorsys/keycloak-config-cli),
 a tool to configure Keycloak.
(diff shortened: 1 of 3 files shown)


kfox1111:
Thanks for the pr. :)

I like the idea of switching to pems from the java trust store.

I think there is one problem with the pr though.

The example should be able to refresh certificates when spire issues new ones. They expire fairly regularly, so that plumbing still needs to be in place. The spiffe-helper and java-spiffe-helper can do the refresh. I think the spire-agent can not. Could spire-agent be switched out with the spire-helper?

moritzschmitz-oviva:
> Thanks for the pr. :)
> 
> I like the idea of switching to pems from the java trust store.
> 
> I think there is one problem with the pr though.
> 
> The example should be able to refresh certificates when spire issues new ones. They expire fairly regularly, so that plumbing still needs to be in place. The spiffe-helper and java-spiffe-helper can do the refresh. I think the spire-agent can not. Could spire-agent be switched out with the spire-helper?

For this example Keycloak itself most likely is the problem. The certs get loaded at startup and are not reloaded during runtime.

kfox1111:
Yeah, there definitely is an issue in Keycloak: https://github.com/keycloak/keycloak/issues/26524

But for the example, we should probably give the recommendation as proper as possible, with rotation working, so when its fixed in Keycloak, those that copied the example don't have to go back and figure out how to fix it.

faisal-memon:
@moritzschmitz-oviva Just wanted to check in if youre still working on this?

faisal-memon:
@moritzschmitz-oviva Just checking in again if you are interested in finishing off this pr?

moritzschmitz-oviva:
> @moritzschmitz-oviva Just checking in again if you are interested in finishing off this pr?

Hi @faisal-memon not at the moment, no. Will probably have a look later if nobody else is interested.

faisal-memon:
Closing due to lack of movement. @moritzschmitz-oviva If you find the time to finish this off please re open. We would appreciate the contribution.

PR #388, opened 2024-06-23
Fix format for ignoreNamespaces

mrsabath:
We missed another place to fix the format of the ignoreNamespaces


kfox1111 (approved):
LGTM. Thanks! :)

PR #392, opened 2024-07-09
Add connect by hostname to agent configmap

tuxotron:
When setting the kubeletConnectByHostname to "true", the charts update the agent daemonset to define the MY_NODE_NAME environment variable, but it doesn't set the "node_name_env" setting in the WorkloadAttestor:k8s plugin, therefore the agent continues trying to connect to the kubelet using the localhost interface

### DIFF
--- a/charts/spire/charts/spire-agent/templates/configmap.yaml
+++ b/charts/spire/charts/spire-agent/templates/configmap.yaml
@@ -116,6 +116,9 @@ plugins:
         disable_container_selectors: {{ .Values.workloadAttestors.k8s.disableContainerSelectors }}
         use_new_container_locator: {{ .Values.workloadAttestors.k8s.useNewContainerLocator }}
         verbose_container_locator_logs: {{ .Values.workloadAttestors.k8s.verboseContainerLocatorLogs }}
+        {{- if eq (include "spire-agent.connect-by-hostname" .) "true" }}
+        node_name_env: "MY_NODE_NAME"
+        {{- end }}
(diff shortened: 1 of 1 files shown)


charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -116,6 +116,9 @@ plugins:
         disable_container_selectors: {{ .Values.workloadAttestors.k8s.disableContainerSelectors }}
         use_new_container_locator: {{ .Values.workloadAttestors.k8s.useNewContainerLocator }}
         verbose_container_locator_logs: {{ .Values.workloadAttestors.k8s.verboseContainerLocatorLogs }}
+        {{- if eq (include "spire-agent.connect-by-hostname" .) "true" }}
+        node_name_env: "MY_NODE_NAME"
+        {{- end }}
```
kfox1111, line 121: According to https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md
node_name_env is defaulted to MY_NODE_NAME when unset.

So, I think this is a  NOOP.

But having it explicit is always nice.

I'm ok with merging this.

mrsabath: I tested this change on Openshift cluster and it works fine.

mrsabath (approved):
LGTM!

kfox1111 (approved):
LGTM. Thanks! :)

PR #407, opened 2024-08-05
Should be `admin_socket_path` not `admin_socket_dir`

bleggett:
AFAICT this is just wrong and should be `admin_socket_path` - the `SPIRE` agent only has references to the latter, and using `dir` results in SPIRE agent config validation errors.


kfox1111 (approved):
LGTM. Thanks! :)

Issue #240, opened 2024-02-07
Provide a way to validate kubelet identity on AKS


orltom:
Hello everyone,

We are encountering an issue with SPIRE while utilizing it via the Helm Chart on Azure AKS, specifically when the `skipKubeletVerification` flag is disabled.

The error we're encountering is as follows:
```
time="2024-02-06T17:28:29Z" level=error msg="Failed to collect all selectors for PID" error="workload attestor \"k8s\" failed: rpc error: code = Internal desc = workloadattestor(k8s): unable to perform request: Get \"https://127.0.0.1:10250/pods\": x509: certificate signed by unknown authority" pid=8747 subsystem_name=workload_attestor
```

It appears that Azure generates separate **self-signed** certificates for the kubelet and kube-api server. Here are some details regarding the certificates:

#### Kube API Certificate
```
> kubectl config view --raw --minify --flatten -o jsonpath='{.clusters[0].cluster.certificate-authority-data}' | base64 --decode | openssl x509 -text -noout | head -n12

Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            46:01:92:5f:0a:b4:5e:18:2a:3b:30:28:20:79:d0:24
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=ca
        Validity
            Not Before: Feb  1 13:35:35 2024 GMT
            Not After : Feb  1 13:45:35 2054 GMT
        Subject: CN=ca
```

#### Spire Agent Certificate 
via Service Account and mounted on /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```
> kubectl get configmaps -n spire kube-root-ca.crt -o json | jq -r '.data[]' |  openssl x509 -text -noout | head -n12

Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            46:01:92:5f:0a:b4:5e:18:2a:3b:30:28:20:79:d0:24
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=ca
        Validity
            Not Before: Feb  1 13:35:35 2024 GMT
            Not After : Feb  1 13:45:35 2054 GMT
        Subject: CN=ca
        Subject Public Key Info:
```

#### Kubelet Certificate
via  Ephemeral Container and OpenSSL CMD
```
>> kubectl debug -it -n spire spire-agent-XYZ --image=ubuntu --target=spire-agent
[...]

Issue #56, opened 2023-10-24
Custom CA


kfox1111:
Until k8s has a native way of adding CA's to the containers chain of trust, we need a way to do it on the spire containers anyway.

marcofranssen:
Could you clarify this issue a bit further @kfox1111? It lacks context for anyone not been part of the discussion.

## My assumption on what I think this is about as a solution:

We are doing this using cert-manager as a upstream ca to have that shared root of trust at the cluster level using a selfsigned CA on cert-manager level.

If you want trust beyond the cluster you could configure cert-manager to utilize PCA on Amazon e.g. to have a CA that is trusted beyond the cluster. Using cert-manager you can make any intermediate CAs on the cluster level, we do this for various components that need to manage their own certificates e.g. spire-ca.

What we can do is create an example of that if that is what this issue is about.

kfox1111:
Other direction. When an organization has their own self signed CA, often for internal use, spire needs to talk to servers with that signature. For example, spire -> mysql, or spire -> federated http based endpoint. This means spire needs to be able to trust the CA for that organization. Currently thats only possible if the root chain of trust within the containers includes the orgs CA. This is currently not configurable.

There is not currently a native way to do this in k8s. It's being worked on though: https://github.com/kubernetes/enhancements/issues/3257

There is a cert-manager subproject that could help in the mean time: https://cert-manager.io/docs/trust/trust-manager/ we could explore if we think that we want to add a dependency for this use case.

kfox1111:
It looks like trust-manager just sets up the configmap though with the whole ca bundle. Its still left up to the user how to inject it into the containers. So we still likely need to add a feature to the chart in that case.

marcofranssen:
> It looks like trust-manager just sets up the configmap though with the whole ca bundle. Its still left up to the user how to inject it into the containers. So we still likely need to add a feature to the chart in that case.

You might be able to use this CA cert to create another cert-manager issuer based on that CA to do the Spire usecase.

kfox1111:
It has nothing to do with issuers. it has to do with the container root chain of trust.

marcofranssen:
> It has nothing to do with issuers. it has to do with the container root chain of trust.

As a workarround you can use the cert-manager issuers to get the secrets created for each container until there is better official support in k8s.

kfox1111:
how do you use cert manager to inject your corperate ca into the spire-server container?

cccsss01:
Happy to look into this.

[...]

PR #415, opened 2024-08-09
Added support for specifying Kubelet CA Path

sabre1041:
Added support for specifying Kubelet CA Path


faisal-memon (approved):
Thanks @sabre1041

charts/spire/charts/spire-agent/README.md
```diff
@@ -88,6 +88,7 @@ A Helm chart to install the SPIRE agent.
...
 | `workloadAttestors.k8s.enabled`                     | Enables the Kubernetes workload attestor                                                                                                                                                          | `true`                                                                           |
 | `workloadAttestors.k8s.skipKubeletVerification`     | If true, kubelet certificate verification is skipped                                                                                                                                              | `true`                                                                           |
+| `workloadAttestors.k8s.kubeletCaPath`               | Path to the file containing CA certificates used to verify the kubelet certificate                                                                                                                | `/run/secrets/kubernetes.io/serviceaccount/ca.crt`                               |
```
kfox1111, line 91: This doesn't follow the documented naming convention.
It should be kubeletCAPath

I'm not seeing where it is being mounted into the container too. Are we messing some code?

kfox1111 (commented):
Thanks for the contribution.

A couple of issues

kfox1111:
This is related to some of the discussion that was happening in https://github.com/spiffe/helm-charts-hardened/issues/56

specifically around, how do we find and mount the ca the kubelet is using into the spire agent. Most of the time, it is not the same CA used for the apiserver.

charts/spire/charts/spire-agent/README.md
kfox1111: Oh, its just assuming that its already there, and its defaulting to the apiserver's CA. In vanilla Kubernetes, that is not the case. :/

kfox1111:
Oh. Found the issue I was looking for:
https://github.com/spiffe/helm-charts-hardened/issues/240

Has a lot of details around a proposed solution

kfox1111:
https://github.com/spiffe/spire/issues/5370 is related.

kfox1111:
I'm thinking, we may not want to allow setting kubelet_ca_path directly at all.
I can think of a few different ways to set it (Types):

KubeletCertificate:
needs a path on disk to the kubelet certificate. On most clusters, its self signed. But it will be just fine to check the tcp connection against the cert kubelet is using itself to serve with.

APIServerSigned:
use the CA provided in /run/secrets/kubernetes.io/serviceaccount/ca.crt.

Specified:
  here's the ca to use as a blob of text

And maybe an 'Auto" or some kind, if the default for vanilla might be different then the default for openshift.

On openshift, are the nodes all self signed by default like in vanilla?

Issue #420, opened 2024-08-12
spire-server v.1.10.0 breaks default chart.


massaox:
Hi,

I have noticed that now the chart `0.21.1` uses the spire-server  `1.10.0` which is set to run as the user 1000. Now if you deploy the chart with the default values you will get this:

```
time="2024-08-12T12:20:34Z" level=warning msg="Current umask 0022 is too permissive; setting umask 0027"
time="2024-08-12T12:20:34Z" level=info msg=Configured admin_ids="[]" data_dir=/run/spire/data launch_log_level=info
time="2024-08-12T12:20:34Z" level=info msg="Opening SQL database" db_type=sqlite3 subsystem_name=sql
time="2024-08-12T12:20:34Z" level=error msg="Fatal run error" error="datastore-sql: unable to open database file: no such file or directory"
time="2024-08-12T12:20:34Z" level=error msg="Server crashed" error="datastore-sql: unable to open database file: no such file or directory"
```

I had to add the following to the values.yaml to make it work.

```
spire-server:
  podSecurityContext:
    fsGroup: 1000
```

Chart version:

```
helm list -A
NAME      	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART           	APP VERSION
spire     	spire    	1       	2024-08-12 14:32:42.985943129 +0000 UTC	deployed	spire-0.21.1    	1.10.0
```

kfox1111:
Hmm... Its passing our testing. How did you deploy it? Is it a fresh install, or is it an upgrade of an existing deployment?

kfox1111:
Filed https://github.com/spiffe/helm-charts-hardened/issues/421

massaox:
Hi. It was a fresh installed, however,  I have just tried again on a fresh GKE cluster. This are the exact command I run:

```
helm upgrade --install --create-namespace -n spire spire-crds spire-crds  --repo https://spiffe.github.io/helm-charts-hardened/
helm upgrade --install -n spire spire spire --repo https://spiffe.github.io/helm-charts-hardened/
```

And got the same error as before:

```
$ kubectl -n spire logs spire-server-0
time="2024-08-13T06:06:46Z" level=warning msg="Current umask 0022 is too permissive; setting umask 0027"
time="2024-08-13T06:06:46Z" level=info msg=Configured admin_ids="[]" data_dir=/run/spire/data launch_log_level=info
[...]

PR #423, opened 2024-08-13
Fix permission issue with 1.10 and defaults when upgraded

kfox1111:
Fixes: https://github.com/spiffe/helm-charts-hardened/issues/420


charts/spire/charts/spire-server/values.yaml
```diff
@@ -1002,6 +1002,22 @@ experimental:
   ## @param experimental.featureFlags [array] List of developer feature flags
   featureFlags: []
 
+# When upgrading from a previous version, sometimes we need to change the user and requires updating permissions.
```
faisal-memon, line 1005: nit: this should go above experimental

kfox1111: Done

faisal-memon (approved):

PR #426, opened 2024-08-14
Enable more forms of kubelet certificate verification

kfox1111:
fixes: https://github.com/spiffe/helm-charts-hardened/issues/240


charts/spire/README.md
```diff
@@ -88,6 +88,12 @@ kubectl delete crds clusterfederatedtrustdomains.spire.spiffe.io clusterspiffeid
 
 We only support upgrading one major/minor version at a time. Version skipping isn't supported. Please see <https://spiffe.io/docs/latest/spire-helm-charts-hardened-about/upgrading/> for details.
 
+### 0.22.X
+
+In previous versions, the setting spire-agent.workloadAttestors.k8s.skipKubeletVerification was set to false by default. Starting in 0.22.x, we removed that setting and replaced it with
```
kfox1111, line 93:
```suggestion
In previous versions, the setting spire-agent.workloadAttestors.k8s.skipKubeletVerification was set to true by default. Starting in 0.22.x, we removed that setting and replaced it with
```

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -106,10 +106,12 @@ plugins:
...
-        # Defaults to the secure kubelet port by default.
-        # Minikube does not have a cert in the cluster CA bundle that
-        # can authenticate the kubelet cert, so skip validation.
-        skip_kubelet_verification: {{ .Values.workloadAttestors.k8s.skipKubeletVerification }}
+        {{- if eq .Values.workloadAttestors.k8s.verification.type "hostCert" }}
+        kubelet_ca_path: /hostCert/kubelet.crt
+        {{- else if eq .Values.workloadAttestors.k8s.verification.type "apiServerCA" }}
+        kubelet_ca_path: /run/secrets/kubernetes.io/serviceaccount/ca.crt
+        {{- end }}
+        skip_kubelet_verification: {{ eq  .Values.workloadAttestors.k8s.verification.type "skip" }}
```
kfox1111, line 114:
```suggestion
        skip_kubelet_verification: {{ eq .Values.workloadAttestors.k8s.verification.type "skip" }}
```

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -206,8 +206,14 @@ workloadAttestors:
...
     ## @param workloadAttestors.k8s.enabled Enables the Kubernetes workload attestor
     enabled: true
-    ## @param workloadAttestors.k8s.skipKubeletVerification If true, kubelet certificate verification is skipped
-    skipKubeletVerification: true
+    verification:
+      ## @param workloadAttestors.k8s.verification.type What kind of verification to do against kubelet. Valid options are [hostCert, apiServerCA, skip]
+      type: hostCert
+      hostCert:
+        ## @param workloadAttestors.k8s.verification.hostCert.basePath Path where kubelet places its certificates
+        basePath: /var/lib/kubelet/pki
```
faisal-memon, line 214: Would be great if we can pull this off. Im worried there will be a lot of variance where this base path is depending on the platform. Whats your experience been with that?

sabre1041 (changes requested):
This is looking very promising. Added a few comments and suggestions

charts/spire/charts/spire-agent/values.yaml
kfox1111: So far, not a lot. I'm sure some distros will use a different value. But we have the same issue with the csi driver too.

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -74,6 +74,36 @@ spec:
         {{- toYaml .Values.hostAliases | nindent 8 }}
       {{- end }}
       initContainers:
+        {{- if eq .Values.workloadAttestors.k8s.verification.type "hostCert" }}
+        - name: gather-host-cert
+          securityContext:
+            {{- $mainSecurityContext | toYaml | nindent 12 }}
+          #FIXME socketAlternate
+          #image: {{ template "spire-lib.image" (dict "image" .Values.socketAlternate.image "global" .Values.global) }}
+          image: chainguard/min-toolkit-debug:latest@sha256:eb1c3f94b5a2303d74a804c1bfd5adc9c98e55e344bbe1adcd6522d6e80d5733
```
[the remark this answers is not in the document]
kfox1111: yeah. thats the #FIXME. just wanted to get something testable going before polishing.

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -74,6 +74,36 @@ spec:
...
+          imagePullPolicy: {{ .Values.socketAlternate.image.pullPolicy }}
+          command: ["bash", "-xec"]
+          args:
+            - |
+              {{- if ne .Values.workloadAttestors.k8s.verification.hostCert.fileName "" }}
+              openssl x509 -in /host/{{ .Values.workloadAttestors.k8s.verification.hostCert.fileName | quote }} -out /hostCert/kubelet.crt
```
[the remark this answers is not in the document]
kfox1111: Its reading through the symlink, which I think should work.

Say we take the openshift layout as an example.

/host here would be mounted from /var/lib/kubelet/pki on the host.

so, if fileName = kubelet-server-current.pem and kubelet-server-current.pem is symlinked to kubelet-server-12345.pem in the same directory, it would look at  /var/lib/kubelet/pki/kubelet-server-current.pem on the host, then redirect to /var/lib/kubelet/pki/kubelet-server-12345.pem and pull out the cert into the emptydir.

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -74,6 +74,36 @@ spec:
...
+                echo Could not find certificate.
+                exit 1
+              fi
+              {{- end }}
+              chmod 644 /hostCert/kubelet.crt
+          volumeMounts:
+          - name: host-cert-isolated
+            mountPath: /hostCert
+          - name: host-cert
+            mountPath: /host
```
kfox1111, line 105:
```suggestion
            mountPath: /host
            readOnly: true
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -220,6 +250,10 @@ spec:
               readOnly: false
             - name: spire-token
               mountPath: /var/run/secrets/tokens
+            {{- if eq .Values.workloadAttestors.k8s.verification.type "hostCert" }}
+            - name: host-cert-isolated
+              mountPath: /hostCert
```
kfox1111, line 255:
```suggestion
              mountPath: /hostCert
              readOnly: true
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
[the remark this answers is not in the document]
sabre1041: it does. it actually tries to assume that its part of the container file system. When i changed from `/host` to `/var/lib/kubelet/pki` it worked without issue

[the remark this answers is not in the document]
kfox1111: oh. is the symlink a full filesystem path?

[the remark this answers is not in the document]
sabre1041: Yup

```
kubelet-server-current.pem -> /var/lib/kubelet/pki/kubelet-server-2024-08-13-12-27-02.pem
```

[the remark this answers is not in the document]
kfox1111: Ah. I see. ok. so, we should probably just mount it in the container at the same place its pulled off the host.

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -74,6 +74,37 @@ spec:
...
+            {{- $mainSecurityContext | toYaml | nindent 12 }}
+          #FIXME socketAlternate
+          #image: {{ template "spire-lib.image" (dict "image" .Values.socketAlternate.image "global" .Values.global) }}
+          image: chainguard/min-toolkit-debug:latest@sha256:eb1c3f94b5a2303d74a804c1bfd5adc9c98e55e344bbe1adcd6522d6e80d5733
+          imagePullPolicy: {{ .Values.socketAlternate.image.pullPolicy }}
+          command: ["bash", "-xec"]
+          args:
+            - |
+              {{- if ne .Values.workloadAttestors.k8s.verification.hostCert.fileName "" }}
+              openssl x509 -in /host/{{ .Values.workloadAttestors.k8s.verification.hostCert.fileName | quote }} -out /hostCert/kubelet.crt
```
kfox1111, line 89:
```suggestion
              openssl x509 -in "{{ .Values.workloadAttestors.k8s.verification.hostCert.basePath }}/{{ .Values.workloadAttestors.k8s.verification.hostCert.fileName }}" -out /hostCert/kubelet.crt
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -74,6 +74,37 @@ spec:
...
+              {{- else }}
+              if [ -f /host/kubelet-server-current.pem ]; then
+                openssl x509 -in /host/kubelet-server-current.pem -out /hostCert/kubelet.crt
+              elif [ -f /host/kubelet.crt ]; then
+                openssl x509 -in /host/kubelet.crt -out /hostCert/kubelet.crt
```
kfox1111, line 94:
```suggestion
              if [ -f "{{ .Values.workloadAttestors.k8s.verification.hostCert.basePath }}/kubelet-server-current.pem" ]; then
                openssl x509 -in "{{ .Values.workloadAttestors.k8s.verification.hostCert.basePath }}/kubelet-server-current.pem" -out /hostCert/kubelet.crt
              elif [ -f "{{ .Values.workloadAttestors.k8s.verification.hostCert.basePath }}/kubelet.crt" ]; then
                openssl x509 -in "{{ .Values.workloadAttestors.k8s.verification.hostCert.basePath }}/kubelet.crt" -out /hostCert/kubelet.crt
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -74,6 +74,37 @@ spec:
...
+                echo Could not find certificate.
+                exit 1
+              fi
+              {{- end }}
+              chmod 644 /hostCert/kubelet.crt
+          volumeMounts:
+          - name: host-cert-isolated
+            mountPath: /hostCert
+          - name: host-cert
+            mountPath: /host
```
kfox1111, line 105:
```suggestion
            mountPath: {{ .Values.workloadAttestors.k8s.verification.hostCert.basePath | quote }}
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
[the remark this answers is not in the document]
kfox1111: Fixed

sabre1041 (approved):
This looks great @kfox1111

charts/spire/README.md
```diff
@@ -88,6 +88,12 @@ kubectl delete crds clusterfederatedtrustdomains.spire.spiffe.io clusterspiffeid
 
 We only support upgrading one major/minor version at a time. Version skipping isn't supported. Please see <https://spiffe.io/docs/latest/spire-helm-charts-hardened-about/upgrading/> for details.
 
+### 0.22.X
```
kfox1111, line 91:
```suggestion
### 0.23.X
```

charts/spire/README.md
```diff
@@ -88,6 +88,12 @@ kubectl delete crds clusterfederatedtrustdomains.spire.spiffe.io clusterspiffeid
...
+
+In previous versions, the setting spire-agent.workloadAttestors.k8s.skipKubeletVerification was set to true by default. Starting in 0.22.x, we removed that setting and replaced it with
```
kfox1111, line 93:
```suggestion
In previous versions, the setting spire-agent.workloadAttestors.k8s.skipKubeletVerification was set to true by default. Starting in 0.23.x, we removed that setting and replaced it with
```

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -206,8 +206,14 @@ workloadAttestors:
   k8s:
     ## @param workloadAttestors.k8s.enabled Enables the Kubernetes workload attestor
     enabled: true
-    ## @param workloadAttestors.k8s.skipKubeletVerification If true, kubelet certificate verification is skipped
-    skipKubeletVerification: true
+    verification:
+      ## @param workloadAttestors.k8s.verification.type What kind of verification to do against kubelet. Valid options are [hostCert, apiServerCA, skip]
+      type: hostCert
```
faisal-memon, line 211: @kfox1111 I like this changeset but Im concerned about this verification not working with some environments. Would you be ok with setting this to `skip` as the default for a release or two? Until we can determine what environments this check will or won't work for.

kfox1111: Hmm... We could... But we're going to need to rip the bandaid off at some point. I'd also like to have some time under our belt with it enabled before we cut a 1.0. So, do we think we're say, 4+ major releases out from 1.0? I was thinking we are pretty close to being able to do a 1.0. Like, maybe we could have 1.0 around kubecon time?

faisal-memon (approved):

charts/spire/charts/spire-agent/values.yaml
faisal-memon: Thanks for setting the default to `skip`

Issue #364, opened 2024-05-23
nested tests failure


kfox1111:
Occasionally it looks like the nested checks fail when trying to test that the ingress controller is available. There is a --wait on the helm install of it, but it looks like it may not be waiting until it really is fully working.

kfox1111:
Seems like everything is ready, but curl is failing sometimes. Like, kube-proxy isn't plumbing up the iptables rules quite fast enough. We probably need to put a retry loop around the curl.

PR #432, opened 2024-08-19
Add missing properties required by SecurityContextConstraints CRD

erikgb:
We add some OpenShift proprietary resource CRDs to more standard Kubernetes clusters like k3s/kind for feature branch testing. Our target clusters are OpenShift, so we want to enable the `global.openshift = true` in the Helm chart. This currently fails because some of the SCCs in the chart does not set some properties required by the SCC CRD: https://github.com/openshift/api/blob/2c10e58877296b062ee6fc63e7fda1eafe7d1bdc/payload-manifests/crds/0000_03_config-operator_01_securitycontextconstraints.crd.yaml#L351

There is something funky here, since the SCCs templated by the chart IS accepted by OpenShift - even if the SCC is missing these mandatory fields. But this appears to be an OpenShift issue, and not something I will try to fix....

### DIFF
--- a/charts/spire/charts/spiffe-csi-driver/templates/scc-spiffe-csi-driver.yaml
+++ b/charts/spire/charts/spiffe-csi-driver/templates/scc-spiffe-csi-driver.yaml
@@ -16,6 +16,7 @@ volumes:
   - configmap
   - hostPath
   - secret
+allowedCapabilities: null
 allowHostDirVolumePlugin: true
 allowHostIPC: false
 allowHostNetwork: false
@@ -23,8 +24,11 @@ allowHostPID: false
 allowHostPorts: false
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: true
+defaultAddCapabilities: null
 fsGroup:
   type: RunAsAny
 groups: []
+priority: null
+requiredDropCapabilities: null
 
 {{ end }}
--- a/charts/spire/charts/spiffe-oidc-discovery-provider/templates/scc-spire-oidc-discovery-provider.yaml
+++ b/charts/spire/charts/spiffe-oidc-discovery-provider/templates/scc-spire-oidc-discovery-provider.yaml
@@ -22,6 +22,7 @@ volumes:
   - hostPath
   - projected
   - secret
+allowedCapabilities: null
 allowHostDirVolumePlugin: true
 allowHostIPC: true
 allowHostNetwork: true
@@ -29,9 +30,12 @@ allowHostPID: true
 allowHostPorts: true
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: true
+defaultAddCapabilities: null
 fsGroup:
   type: RunAsAny
 groups: []
+priority: null
+requiredDropCapabilities: null
 seccompProfiles:
   - '*'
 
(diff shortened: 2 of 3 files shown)


sabre1041:
@erikgb what OpenShift environment/version are you testing against?

erikgb:
> @erikgb what OpenShift environment/version are you testing against?

@sabre1041 The SCC CRD is downloaded from the API server in a OCP 4.15.11 cluster (our target environment). For feature branch pipelines in our provisioning projects we use a k3s cluster with selective OpenShift CRDs added. We haven't found any usable ephemeral OpenShift clusters to use.

sabre1041:
> > @erikgb what OpenShift environment/version are you testing against?
> 
> @sabre1041 The SCC CRD is downloaded from the API server in a OCP 4.15.11 cluster (our target environment). For feature branch pipelines in our provisioning projects we use a k3s cluster with selective OpenShift CRDs added. We haven't found any usable ephemeral OpenShift "brand" clusters to use.
> 
> We prefer to avoid configuring our software **for** OpenShift (more Kubernetes standard resources), but with regards to SCCs we have no choice.

I know that OpenShift is most lenient as it relates to CR/schema for CRD's so its not entirely surprising that you are tuning into these challenges when operating in a standard kubernetes environment. Have you tried to install the chart into a fully supported OpenShift environment and does it work as expected?

erikgb:
> Have you tried to install the chart into a fully supported OpenShift environment and does it work as expected?

Yes, as stated in the PR description. But then I am unable to verify changes on feature branches. The CRDs published by OpenShift are obviously wrong and a bug, but I have no influence over what Red Hat does. This PR is a workaround.

kfox1111:
The other route is to fix the CRDS so they are more representative of what OpenShift is going to do. You could just remove the faulty requirements section out of the CRDS before loading it since it seems like OpenShift isn't honoring it? The cluster should then behave more like OpenShift and work with charts other then just this one.

A quick fix may be:
```
curl https://github.com/openshift/api/raw/2c10e58877296b062ee6fc63e7fda1eafe7d1bdc/payload-manifests/crds/0000_03_config-operator_01_securitycontextconstraints.crd.yaml -L \
  | yq e '.spec.versions[0].schema.openAPIV3Schema.required = []' - \
  | kubectl apply -f -
```

erikgb:
> The other route is to fix the CRDS so they are more representative of what OpenShift is going to do. You could just remove the faulty requirements section out of the CRDS before loading it since it seems like OpenShift isn't honoring it?

Yes, that's an option. Adding even more complexity. But IMO being explicit in the SCCs is better - and also aligned with the faulty CRDs published by OpenShift. The problem with this chart specifically is that it's built for OpenShift (`global.openshift: true`). We didn't use this value previously and fixed our OpenShift adaptions using patches. But with the latest release, a problematic initContainer was added - based on a conditional including this property....

sabre1041 (approved):
Given that some of the out of the box OpenShift SCC's also include null values, 👍 

Verified on OpenShift Container Platform 4.15 and KIND

kfox1111:
Ok. Once the DCO is fixed, then we can merge.

erikgb:
> Ok. Once the DCO is fixed, then we can merge.

Thanks @kfox1111! DCO should be OK now.

erikgb:
There seems to be some flakiness in CI. I saw the same job failing on the first run, and appears to be completely unrelated.

kfox1111:
https://github.com/spiffe/helm-charts-hardened/issues/364

kfox1111 (approved):
LGTM. Thanks! :)

PR #448, opened 2024-09-06
Set readOnlyRootFilesystem to false in spire-agent SCC to avoid CLBO

bendikp:
We couldn't get spire-agent running after switching to `openshift: true`. The spire-agent was reporting these errors and got stuck in CrashLoopBackOff.

```
time="2024-09-06T08:32:01Z" level=info msg="Creating spire agent UDS directory" dir=/run/spire/agent-sockets
mkdir /run/spire/agent-sockets: read-only file system
```

After changing the `readOnlyRootFilesystem` to `false` in the spire-agent SCC the spire-agent was able to start up.

### DIFF
--- a/charts/spire/charts/spire-agent/templates/scc-spire-agent.yaml
+++ b/charts/spire/charts/spire-agent/templates/scc-spire-agent.yaml
@@ -3,7 +3,7 @@ apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
 metadata:
   name: {{ include "spire-agent.fullname" . }}
-readOnlyRootFilesystem: true
+readOnlyRootFilesystem: false
 runAsUser:
   type: RunAsAny
 seLinuxContext:


kfox1111:
Hmm... readOnlyRootFilesystem to false I think fixes a symptom and not the cause?

IE, its best to keep ti readonly if we can find a way for it to function that way.

bendikp:
Based on this [documentation](https://github.com/spiffe/spire/blob/67b86c036ac8ca9d490ea4c78cdfa0e63253f93d/doc/spire_agent.md#security-context-constraints) it says that `readOnlyFilesystem` should be set to `false`.

kfox1111:
Based on the conversation I had on slack, The configmap was changed after the chart rendered it, likely introducing the issue, that this pr is trying to workaround. The log posted here is not possible to generate from the chart when configured via values, and involves it not being able to make a path, improperly, within the container.

It will be important to try and configure the chart via vales rather then patching the chart to work around an issue introduced by performing an unsupported operation. If the bug still shows up after that, or other issues arise, we can address them then.

There are folks running on openshift successfully without this patch. I think its likely the modified configmap is the problem after spending a couple hours trying to debug this.

As to whether the option should be read only or not, it should be read only. There is no reason for the root filesystem of the container to need to be read write. We're successfully running in that configuration on other systems.

bendikp:
Okey, thanks. I'll have a look at the conversation on Slack and see if we can make it work without disabling readonly root filesystem.

bendikp:
We've now fixed our spire config, and it works without disabling readonly root filesystem 🥳 

Thanks for your input.

PR #450, opened 2024-09-10
Protects SPIRE Agent's Chart Against YAML Template Injection

nweisenauer-sap:
Hi,

while the SPIRE Agent Helm chart already uses the `quote` and `toYaml` functions to protect itself against YAML injection via its Go Templates in a lot of spots, I decided to put some more effort into it and protect also the rest of the places against injection where those functions were not used, yet.

### Background
Helm uses Go Templates to generate YAML files that define K8S resources. These Helm templates can process values that come from outside the Helm chart itself.
In a complex Infrastructure-as-Code (IaC) landscape, some of the values may come from all sorts of sources. Eventually, some of these sources will be controlled by potential attackers, at least to some degree. Data from such sources is sometimes called "tainted data". Therefore, attackers can use tainted data to inject attack payloads into the values that a Helm chart processes.
On the other hand, Go Templates are plain text templates. Therefore, Helm templates are not aware of YAML syntax at all. This allows attackers to inject attack payloads that cause unintended changes to the generated YAML data, and ultimately to managed K8S resources.

Cheers,
 Nico


marcofranssen:
Looks like a nice improvement, approved the workflow run. 🤞

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -150,7 +150,9 @@ health_checks:
 telemetry:
   - Prometheus:
       - host: "0.0.0.0"
+        {{- if ne (int .port) 0 }}
         port: {{ .Values.telemetry.prometheus.port }}
+        {{- end }}
```
kfox1111, line 155:
```suggestion
        port: {{ .Values.telemetry.prometheus.port | int }}
```

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -168,8 +170,8 @@ telemetry:
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ include "spire-agent.fullname" . }}{{ $nameSuffix }}
-  namespace: {{ include "spire-agent.namespace" . }}
+  name: {{ printf "%s%s" (include "spire-agent.fullname" .) ($nameSuffix) | quote }}
```
kfox1111, line 173:
```suggestion
  name: {{ printf "%s%s" (include "spire-agent.fullname" .) $nameSuffix | quote }}
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -22,16 +22,16 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: {{ include "spire-agent.fullname" . }}{{ $nameSuffix }}
-  namespace: {{ include "spire-agent.namespace" . }}
+  name: {{ printf "%s%s" (include "spire-agent.fullname" .) ($nameSuffix) | quote }}
```
kfox1111, line 25:
```suggestion
  name: {{ printf "%s%s" (include "spire-agent.fullname" .) $nameSuffix | quote }}
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -79,17 +79,17 @@ spec:
...
             {{- $mainSecurityContext | toYaml | nindent 12 }}
           image: {{ template "spire-lib.image" (dict "image" .Values.hostCert.image "global" .Values.global) }}
-          imagePullPolicy: {{ .Values.hostCert.image.pullPolicy }}
+          imagePullPolicy: {{ .Values.hostCert.image.pullPolicy | quote }}
           command: ["bash", "-xc"]
           args:
             - |
               {{- if ne .Values.workloadAttestors.k8s.verification.hostCert.fileName "" }}
-              openssl x509 -in "{{ .Values.workloadAttestors.k8s.verification.hostCert.basePath }}/{{ .Values.workloadAttestors.k8s.verification.hostCert.fileName }}" -out /hostCert/kubelet.crt
+              openssl x509 -in {{ printf "%s/%s" .Values.workloadAttestors.k8s.verification.hostCert.basePath  .Values.workloadAttestors.k8s.verification.hostCert.fileName | quote }} -out /hostCert/kubelet.crt
```
kfox1111, line 87:
```suggestion
              openssl x509 -in {{ printf "%s/%s" .Values.workloadAttestors.k8s.verification.hostCert.basePath .Values.workloadAttestors.k8s.verification.hostCert.fileName | quote }} -out /hostCert/kubelet.crt
```

charts/spire/charts/spire-agent/templates/configmap.yaml
nweisenauer-sap: This suggestion would configure a default port of `0` if someone specifies a port that cannot be cast to int, e.g. the string "8000" instead of number `8000`
I personally prefer to not put the port if someone does not specify an int instead of defaulting it to `0`, what do you think?

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -314,7 +314,11 @@ spec:
         {{- if .Values.sockets.admin.mountOnHost }}
         - name: spire-agent-admin-socket-dir
           hostPath:
-            path: {{ .Values.sockets.hostBasePath }}/{{ if .Values.upstream }}upstream.csi.spiffe.io{{ else }}csi.spiffe.io{{ end }}/admin
+            {{- if .Values.upstream }}
+            path: {{ printf "%s/upstream.csi.spiffe.io/admin" (.Values.sockets.hostBasePath) | quote }}
+            {{- else }}
+            path: {{ printf "%s/csi.spiffe.io/admin" (.Values.sockets.hostBasePath) | quote }}
```
kfox1111, line 320:
```suggestion
            path: {{ printf "%s/upstream.csi.spiffe.io/admin" .Values.sockets.hostBasePath | quote }}
            {{- else }}
            path: {{ printf "%s/csi.spiffe.io/admin" .Values.sockets.hostBasePath | quote }}
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -323,7 +327,11 @@ spec:
         {{- if eq .Values.persistence.type "hostPath" }}
         - name: spire-agent-persistence
           hostPath:
-            path: {{ .Values.persistence.hostPath }}/{{ if .Values.upstream }}upstream.csi.spiffe.io{{ else }}csi.spiffe.io{{ end }}
+            {{- if .Values.upstream }}
+            path: {{ printf "%s/upstream.csi.spiffe.io" (.Values.persistence.hostPath) | quote }}
+            {{- else }}
+            path: {{ printf "%s/csi.spiffe.io" (.Values.persistence.hostPath) | quote }}
```
kfox1111, line 333:
```suggestion
            path: {{ printf "%s/upstream.csi.spiffe.io" .Values.persistence.hostPath | quote }}
            {{- else }}
            path: {{ printf "%s/csi.spiffe.io" .Values.persistence.hostPath | quote }}
```

charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -332,7 +340,7 @@ spec:
         {{- if eq (len .Values.trustBundleURL) 0 }}
         - name: spire-bundle
           configMap:
-            name: {{ include "spire-lib.bundle-configmap" . }}{{ $nameSuffix }}
+            name: {{ printf "%s%s" (include "spire-lib.bundle-configmap" .) ($nameSuffix) | quote }}
```
kfox1111, line 343:
```suggestion
            name: {{ printf "%s%s" (include "spire-lib.bundle-configmap" .) $nameSuffix | quote }}
```

kfox1111 (commented):
Thanks for working on this.

I think the premise that deploying a security sensitive service using sources of config that are not totally under your control is not something we should be encouraging and we need to be careful to claim we protect against.

That being said, the changes look good and does enhance the project though. So, I'm onboard with the changes.

A few changes suggested inline.

charts/spire/charts/spire-agent/templates/configmap.yaml
kfox1111: Could add a second line around it to fail the deploy if it isn't an int?

kfox1111: There are other places in the PR that just cast to int as well. Should there be some consistency there?

nweisenauer-sap: I wanted to be consistent with the existing port validation implemented in the same file [in L87](https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-agent/templates/configmap.yaml#L87) and [in L90](https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-agent/templates/configmap.yaml#L90).
In the daemonset.yaml I used `int` without this surrounding if statement. So for consistency, how should we deal with it? Go with the simple `| int` cast for ports in the configmap.yaml and change L87 and L90, too?

kfox1111: We discussed this at the sync.

The plan:
We're going to add a json schema file for the chart. We will initially set all the properties that need to be integers as integers in it. This means helm will verify all the input values are integers before it hits the template engine, so the templates don't have to do anything to recheck its an integer or cast. We will do that in a separate PR.

So, in this PR, lets:
remove any ` | int`

And the little bit of the pr here that started the conversation about adding the if's here, lets remove those too.

I think the rest of the pr looks good then.

kfox1111: https://github.com/spiffe/helm-charts-hardened/pull/458

nweisenauer-sap: Cool, thanks for going the extra mile 👍

kfox1111 (approved):
LGTM. Thanks! :)

PR #462, opened 2024-09-17
Change curl --caPath to --capath

### LINKED ISSUE #461: Spire Agent kubelet verification: curl: option --caPath: is unknown
It looks like in version `0.23.0` the `spire-agent.workloadAttestors.k8s.verification.type` feature does not work properly. Curl does not have a `--caPath` option. It looks like it's case sensitive and should be `--capath`. Location in the [source code](https://github.com/spiffe/helm-charts-hardened/blob/8832da39b4549c942785255730fe030fad4d81ef/charts/spire/charts/spire-agent/templates/daemonset.yaml#L100)

I could verify that in the container image of the `gather-host-cert` container as follows:
```
docker run cgr.dev/chainguard/bash:latest@sha256:f5c85affd2aa0f55fc1ead7dc07952577ad82741bbbba742ead0fd9dde2de14a "curl --caPath"
curl: option --caPath: is unknown
```
Passing the option all lowercase looks better:
```
docker run cgr.dev/chainguard/bash:latest@sha256:f5c85affd2aa0f55fc1ead7dc07952577ad82741bbbba742ead0fd9dde2de14a "curl --capath"
curl: option --capath: requires parameter
```

faisal-memon:
Changes the curl option to be lower case. the camel case options works on mac but not linux.

fixes #461


kfox1111 (approved):
LGTM. Thanks! :)

PR #463, opened 2024-09-18
Add remaining data store options

kfox1111:
We dont support all the spire server datastore sql plugin options. Add the missing ones.

Before the change, we only support:
```
    ## @param dataStore.sql.databaseType Other supported databases are "postgres" and "mysql"
    databaseType: sqlite3
    ## @param dataStore.sql.databaseName Only used by "postgres" or "mysql"
    databaseName: spire
    ## @param dataStore.sql.host Only used by "postgres" or "mysql"
    host: ""
    ## @param dataStore.sql.port If 0 (default), it will auto set to 5432 for postgres and 3306 for mysql. Only used by those databases.
    port: 0
    ## @param dataStore.sql.username Only used by "postgres" or "mysql"
    username: spire
    ## @param dataStore.sql.password Only used by "postgres" or "mysql"
    password: ""
    ## @param dataStore.sql.options [array] Only used by "postgres" or "mysql"
    options: []

    ## @param dataStore.sql.plugin_data [object] Settings from https://github.com/spiffe/spire/blob/main/doc/plugin_server_datastore_sql.md go in this section
    plugin_data: {}

    ## When an external source creates the secret. The secret should reside in the same namespace as the spire server
    externalSecret:
      ## @param dataStore.sql.externalSecret.enabled Enable external secret for datastore creds
      enabled: false
      ## @param dataStore.sql.externalSecret.name The name of the secret object
      name: ""
      ## @param dataStore.sql.externalSecret.key The key of the secret object whose value is the dataStore.sql password
      key: ""
```


charts/spire/README.md
```diff
@@ -88,6 +88,11 @@ kubectl delete crds clusterfederatedtrustdomains.spire.spiffe.io clusterspiffeid
 
 We only support upgrading one major/minor version at a time. Version skipping isn't supported. Please see <https://spiffe.io/docs/latest/spire-helm-charts-hardened-about/upgrading/> for details.
 
+### 0.24.X
+
+We havd added the remaining options needed for the SPIRE Server sql data store plugin. We have removed spire-server.dataStore.sql.plugin_data as it is no longer needed. If you are using it, please migrate
```
faisal-memon, line 93:
```suggestion
We have added the remaining options needed for the SPIRE Server SQL data store plugin. We have removed `spire-server.dataStore.sql.plugin_data` section as it is no longer needed. If you are using it, please migrate
```

charts/spire/README.md
```diff
@@ -88,6 +88,11 @@ kubectl delete crds clusterfederatedtrustdomains.spire.spiffe.io clusterspiffeid
...
+your settings to the ones under spire-server.dataStore.sql.
```
faisal-memon, line 94:
```suggestion
your settings to the ones under `spire-server.dataStore.sql`.
```

charts/spire/charts/spire-server/values.yaml
```diff
@@ -182,6 +186,37 @@ dataStore:
...
+      ## @param dataStore.sql.readOnly.options [array] Only used by "postgres" or "mysql"
+      options: []
+      ## When an external source creates the secret. The secret should reside in the same namespace as the spire server
+      externalSecret:
+        ## @param dataStore.sql.readOnly.externalSecret.enabled Enable external secret for datastore creds
+        enabled: false
+        ## @param dataStore.sql.readOnly.externalSecret.name The name of the secret object
+        name: ""
+        ## @param dataStore.sql.readOnly.externalSecret.key The key of the secret object whose value is the dataStore.sql password
+        key: ""
```
faisal-memon, line 218: Does the read only need a separate externalSecret option?

kfox1111: The ro password could be in a separate secret. It can be combined too though if you use the same name and different keys.

kfox1111: I double checked.... The bitnami postgresql chart lets you create one user via values and loads the pw into its secret. If a user wants to configure another ro user, then they would have to do so manually and load it into a different secret. So, the functionality is required in at least one configuration. So, I think we want these options.

kfox1111:
Needs support for:
```
connection_string="file:memdb?mode=memory&cache=shared"
```

kfox1111:
> Needs support for:
> 
> ```
> connection_string="file:memdb?mode=memory&cache=shared"
> ```

Moving to separate issue: 
https://github.com/spiffe/helm-charts-hardened/issues/465

edwbuck (commented):
Ideally some document or unit test that ensures the sql plugin is configured to match the examples under https://github.com/spiffe/spire/blob/v1.10.0/doc/plugin_server_datastore_sql.md would be included, to ensure that we have the settings documented to correctly derive the examples.

charts/spire/README.md
```diff
@@ -88,6 +88,11 @@ kubectl delete crds clusterfederatedtrustdomains.spire.spiffe.io clusterspiffeid
 
 We only support upgrading one major/minor version at a time. Version skipping isn't supported. Please see <https://spiffe.io/docs/latest/spire-helm-charts-hardened-about/upgrading/> for details.
 
+### 0.24.X
+
+We have added the remaining options needed for the SPIRE Server SQL data store plugin. We have removed `spire-server.dataStore.sql.plugin_data` section as it is no longer needed. If you are using it, please migrate
```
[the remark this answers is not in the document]
faisal-memon: this is done

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -101,7 +101,22 @@ plugins:
   DataStore:
     sql:
       plugin_data:
-        {{ include "spire-server.datastore-config" . | nindent 10 }}
+        {{ include "spire-server.datastore-config" . | nindent 8 }}
+        {{- if ne .Values.dataStore.sql.rootCAPath "" }}
```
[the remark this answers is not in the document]
faisal-memon: This is done

faisal-memon:
@edwbuck Changes look good to me. Can you give final approval?

faisal-memon (approved):

edwbuck (approved):

PR #476, opened 2024-10-10
Fix cert-manager upstream authority when enabling recommendations

yann-soubeyrand:
The namespace used in the command of the wait init container should be the same as the issuer resource: https://github.com/spiffe/helm-charts-hardened/blob/cecf51c27c9e280330b61dcb970df0272c83ae27/charts/spire/charts/spire-server/templates/issuer.yaml#L50
Else, it doesn’t work when deploying with recommendations enabled.


kfox1111 (approved):
LGTM. Thanks! :)

PR #478, opened 2024-10-14
Agent support for bundle on host path

kfox1111:
When integrating with a SPIRE setup that has trust established from the host, there needs to be a way to pass in the trust bundle from a host file.


charts/spire/charts/spire-agent/values.yaml
```diff
@@ -95,6 +95,8 @@ trustDomain: example.org
 trustBundleURL: ""
 ## @param trustBundleFormat If using trustBundleURL, what format is the url. Choices are "pem" and "spiffe"
 trustBundleFormat: pem
+## @param trustBundleHostPath If set, obtain trust bundle from a file on the host
```
faisal-memon, line 98:
```suggestion
## @param trustBundleHostPath If set, obtain trust bundle from a file on the host instead of from the ConfigMap
```

faisal-memon (approved):

Issue #439, opened 2024-08-23
Validating Admission Policy


kfox1111:
CEL based validating admission policy is stable now (https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/)

we can add a policy that says, only namespace spire-server can use csi driver upstream.csi.spiffe.io

PR #482, opened 2024-10-23
Add validating admission policy to restrict upstream driver access

kfox1111:
Fixes: https://github.com/spiffe/helm-charts-hardened/issues/439

### DIFF
--- a/charts/spire/charts/spiffe-csi-driver/values.yaml
+++ b/charts/spire/charts/spiffe-csi-driver/values.yaml
@@ -60,6 +60,13 @@ nameOverride: ""
 ## @param namespaceOverride Namespace to install spiffe-csi-driver
 namespaceOverride: ""
 
+## @param serverNamespaceOverride Override the namespace that the spire-server is installed into
+serverNamespaceOverride: ""
+
+validatingAdmissionPolicy:
+  ## @param validatingAdmissionPolicy.enabled When auto, an upstream driver, and k8s >= 1.30.0 then enabled = true, otherwise false
+  enabled: auto
+
 ## @param fullnameOverride Full name override for spiffe-csi-driver
 fullnameOverride: ""
 
--- a/.github/workflows/helm-chart-ci-ignore.yaml
+++ b/.github/workflows/helm-chart-ci-ignore.yaml
@@ -30,9 +30,9 @@ jobs:
     strategy:
       matrix:
         k8s:
-          - v1.28.0
-          - v1.27.3
-          - v1.26.6
+          - v1.31.1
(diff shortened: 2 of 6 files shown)


charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -60,6 +60,13 @@ nameOverride: ""
 ## @param namespaceOverride Namespace to install spiffe-csi-driver
 namespaceOverride: ""
 
+## @param serverNamespaceOverride Override the namespace that the spire-server is installed into
+serverNamespaceOverride: ""
+
+validatingAdmissionPolicy:
+  ## @param validatingAdmissionPolicy.enabled When auto, an upstream driver, and k8s >= 1.30.0 then enabled = true, otherwise false
```
faisal-memon, line 67: Im not understanding the upstream driver part of this comment. Seems like some qualifier is missing.

kfox1111: {{- $upstream := eq .Values.pluginName "upstream.csi.spiffe.io" }}

Could word that explicitly. something like this maybe?:
```
  ## @param validatingAdmissionPolicy.enabled When auto, pluginName == "upstream.csi.spiffe.io", and k8s >= 1.30.0 then enabled = true, otherwise false
```

faisal-memon: Yes thats better

kfox1111: Sounds good. Merged.

charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -60,6 +60,13 @@ nameOverride: ""
 ## @param namespaceOverride Namespace to install spiffe-csi-driver
 namespaceOverride: ""
 
+## @param serverNamespaceOverride Override the namespace that the spire-server is installed into
+serverNamespaceOverride: ""
+
+validatingAdmissionPolicy:
+  ## @param validatingAdmissionPolicy.enabled When auto, pluginName == "upstream.csi.spiffe.io", and k8s >= 1.30.0 then enabled = true, otherwise false
```
faisal-memon, line 67:
```suggestion
  ## @param validatingAdmissionPolicy.enabled When set to auto, the validatingAdmissionPolicy will be enabled when the pluginName == "upstream.csi.spiffe.io" and k8s >= 1.30.0. Valid options are [auto, true, false]
```

faisal-memon (approved):

PR #491, opened 2024-11-01
Fix GCS Bundle endpoint format variable

MattiasGees:
The GCS Bundle endpoint configuration was pointing to the S3 Format variable instead of the GCS one.


marcofranssen (approved):
LGTM 🚀 

See https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-server/values.yaml#L929

PR #534, opened 2025-03-04
Fixes indentation of sync_interval value inside spire-agent configmap

nweisenauer-sap:
The indentation of the `sync_interval` was wrong, so it could not be set. This PR fixes the issue.


kfox1111 (approved):
LGTM. Thanks! :)

Issue #47, opened 2023-10-18
Add support for SPIRE rotating webhook ca bundles


faisal-memon:
Documented here: https://github.com/spiffe/spire/blob/main/doc/plugin_server_notifier_k8sbundle.md#:~:text=webhook_label,to%20true.

PR #536, opened 2025-03-09
Add values to configure apiServiceLabel and webhookLabel

tamalsaha:
Fixes https://github.com/spiffe/helm-charts-hardened/issues/47


charts/spire/charts/spire-server/values.yaml
```diff
@@ -487,6 +487,10 @@ notifier:
     enabled: true
     ## @param notifier.k8sBundle.namespace Namespace to push the bundle into, if blank will default to SPIRE Server namespace
     namespace: ""
+    ## @param notifier.k8sBundle.apiServiceLabel If set, rotate the CA Bundle in API services with this label set to true.
+    apiServiceLabel: ""
```
kfox1111, line 491: Can we make these bools, defaulted to false?

kfox1111 (commented):
Looking good. Thanks for the fix. One issue inline.

charts/spire/charts/spire-server/values.yaml
tamalsaha: These are the "key" part of the label. Say 

`apiServiceLabel: "example.com/spiffe"`. Then the actual label on an ApiService has to be `"example.com/spiffe": "true"`.

kfox1111: OHH. That makes more sense.

When webhooks are deployed by the chart, does it make sense to always do the rotation automatically?

tamalsaha: I am too new to spiffe to answer this question. I see that networkservicemesh project uses this features using their static manifest. I am just trying to reproduce the same behavior using helm chart.

kfox1111: Ok, thanks. I'll ask around. Will try and get back to you soon.

kfox1111: Ok, so it sounds like those options were added primarily for the predecessor of spire-controller-manager. Spire-controller-manager will do it itself, so having the options aren't bad, but not needed for proper functioning of the spire-controller-manager. So, I think we're good the way the PR is.

kfox1111 (approved):
LGTM. Thanks! :)

PR #547, opened 2025-03-27
Add new config.jwksUri option to oidc-discovery chart

RedneckEdamame:
Add an option to configure JWKS URI. An example of why users may want to do this if they are running on a non-standard port. This option is already supported in the upstream spiffe/spire project.


kfox1111 (approved):
LGTM. Thanks! :)

Issue #554, opened 2025-04-07
spiffe-csi driver update


kfox1111:
There are newer versions at https://github.com/spiffe/spiffe-csi/releases

PR #555, opened 2025-04-07
Update csi driver

kfox1111:
Fixes: #554


erikgb:
LGTM, but are we able to add automation to avoid doing this manually when newer releases are available?

kfox1111:
> LGTM, but are we able to add automation to avoid doing this manually when newer releases are available?

only a matter of code... probably its own task.

faisal-memon (approved):

PR #563, opened 2025-04-18
Update spire-controller-manager to 0.6.2 and add its staticManifest support

kfox1111:
Static manifest mode allows ClusterStaticEntries and ClusterFederatedTrustBundles to be managed by the chart without needing the Kubernetes CRDs, or loaded into an externally managed configmap and synced into the server.

This enables installing and managing spire-servers on clusters where you only have admin access to a single namespace, not cluster-admin to the entire cluster.


charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
```diff
@@ -75,7 +75,11 @@
     {{- end }}
   env:
     - name: ENABLE_WEBHOOKS
+    {{- if eq .Values.controllerManager.staticManifestMode "off" }}
       value: {{ .webhooksEnabled | toString | quote }}
+    {{- else }}
+      value: "false"
```
faisal-memon, line 81: The static manifest mode doesnt need webhooks?

kfox1111: no. It reads the static manifests right from the file system. In this case, provided by a configmap.

faisal-memon: It can only do one at a time?

kfox1111: Correct. it either pulls the yaml objects from a local directory (static manifests) or from k8s

Issue #568, opened 2025-04-23
Invalid image name when using digests in spire Helm Chart


root30:
The `spire-agent` Helm chart has a bug when using only a digest as the tag in the `values.yaml` i.e.
```
...
spire-agent:
  hostCert:
    image:
      tag: sha256:8699d1707c16f2e05e321d19904652c16090b5819d657a91efd051d437f1b7dd
...
```
This produces invalid images names with double `/` between registry and repo like
`cgr.dev//chainguard/min-toolkit-debug@sha256:8699d1707c16f2e05e321d19904652c16090b5819d657a91efd051d437f1b7dd `

The reason can be found in the `spire-lib.image` template funtion on [L56](https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-lib/templates/_helpers.tpl#L56) in the `_helpers.tpl` of the `spire-lib`.
The `$registry` variable will always be suffixed with `/`  by the `spire-lib.registry` template funtion and therefore the format string in L56 should be changed to `"%s%s@%s"` as in L62, L64 or L66.

PR #569, opened 2025-04-23
Fix invalid image name for digest in template function of `spire-lib`

root30:
Fixes [Invalid image name when using digests in spire Helm Chart #568](https://github.com/spiffe/helm-charts-hardened/issues/568)


kfox1111 (approved):
LGTM. Thanks! :)

Issue #465, opened 2024-09-19
sqlite in memory support


kfox1111:
We  don't have any way to generate a connection string that looks like:
```
connection_string="file:memdb?mode=memory&cache=shared"
```

PR #576, opened 2025-05-05
Dynamically build the connection_string with options for the spire server when spire-server.dataStore.sql.databaseType == sqlite3

daescha:
Fixes: #465


daescha:
@kfox1111 okay, I think I fixed it now

daescha:
so we need options for:

- file
- cache
- immutable
- nolock
- mode
- psow
- modeof
- vfs

based on https://www.sqlite.org/uri.html. is that correct?

daescha:
@kfox1111 I think the options are how you want now. One thing I was unsure on is the file for the connection string. I made a new option `file` for sqlite only, not sure if thats best. 

also, its not very clear to me how to use options. I had to plug all the templates into a helm playground to figure out the correct format for items in the array. Perhaps the docs can mention what the proper format is supposed to be i.e. `items in the list should be objects of form {"key": "", value: ""}`

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -194,7 +208,8 @@ Create the name of the service account to use
 {{- $ropw := "" }}
 {{- if eq .Values.dataStore.sql.databaseType "sqlite3" }}
   {{- $_ := set $config "database_type" "sqlite3" }}
-  {{- $_ := set $config "connection_string" "/run/spire/data/datastore.sqlite3" }}
+  {{- $query := include "spire-server.config-sqlite-query" .Values.dataStore.sql.options }}
+  {{- $_ := set $config "connection_string" (printf "file:%s%s" (default "/run/spire/data/datastore.sqlite3" .Values.dataStore.sql.file) $query) }}
```
daescha, line 212:
```suggestion
  {{- $_ := set $config "connection_string" (printf "%s%s" (default "/run/spire/data/datastore.sqlite3" .Values.dataStore.sql.file) $query) }}
```

should the `file` scheme be automatically inserted like I have it or left to users to add? If you don't include the scheme, would that cause an error if you use query options?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -167,7 +167,9 @@ dataStore:
     username: spire
     ## @param dataStore.sql.password Only used when type != "sqlite3"
     password: ""
-    ## @param dataStore.sql.options [array] Only used when type != "sqlite3"
+    ## @param dataStore.sql.file Only used when type == "sqlite3"
+    file: ""
+    ## @param dataStore.sql.options [array]
```
daescha, line 172:
```suggestion
    ## @param dataStore.sql.options [array] takes an array of objects of form {<key>: <value>} to use when building the database connection string
```

idea for improvement of options documentation (obviously need to regen the readme if accepted)

charts/spire/charts/spire-server/values.yaml
```diff
@@ -167,7 +167,9 @@ dataStore:
     username: spire
     ## @param dataStore.sql.password Only used when type != "sqlite3"
     password: ""
-    ## @param dataStore.sql.options [array] Only used when type != "sqlite3"
+    ## @param dataStore.sql.file Only used when type == "sqlite3"
```
daescha, line 170:
```suggestion
    ## @param dataStore.sql.file Data source file. Only used when type == "sqlite3"
```

add extra info about the file parameter (obviously need to regen the readme if accepted)

daescha (commented):
hi @kfox1111, I added some changes I was unsure about as review comments

faisal-memon (changes requested):
Thanks for submitting this PR

charts/spire/charts/spire-server/README.md
```diff
@@ -131,7 +131,8 @@ In order to run Tornjak with simple HTTP Connection only, make sure you don't cr
...
+| `dataStore.sql.file`                                                                         | Only used when type == "sqlite3"                                                                                                                                                                                                                                                                                                                                     | `""`                                                                                           |
+| `dataStore.sql.options`                                                                      |                                                                                                                                                                                                                                                                                                                                                                      | `[]`                                                                                           |
```
faisal-memon, line 135: Is this `options` for sqlite only?

charts/spire/charts/spire-server/templates/_helpers.tpl
faisal-memon: I think you should set the default in the values file

charts/spire/charts/spire-server/README.md
daescha: previously options was for all types other than sqlite3. this PR makes it so it now applies to sqlite as well, so no restrictions anymore.

kfox1111:
Looks ok to me. @faisal-memon ?

faisal-memon (approved):

faisal-memon:
Thanks @daescha for making the requested changes.

Issue #584, opened 2025-05-18
spire credentialcomposer cel support


kfox1111:
Add support for https://github.com/spiffe/spire-credentialcomposer-cel

kfox1111:
Useful for stuff like:
https://github.com/spiffe/club-zero/issues/7
and 
https://github.com/spiffe/club-zero/issues/9

PR #587, opened 2025-05-19
Add spire-credentialcomposer-cel support

kfox1111:
fixes: https://github.com/spiffe/helm-charts-hardened/issues/584


charts/spire/charts/spire-server/templates/server-resource.yaml
```diff
@@ -121,8 +121,41 @@ spec:
...
+        - name: init-cel
+          securityContext:
+            {{- include "spire-lib.securitycontext" . | nindent 12 }}
+          image: {{ template "spire-lib.image" (dict "appVersion" $.Chart.AppVersion "image" .Values.tools.busybox.image "global" .Values.global) }}
+          command:
+            - busybox
+            - sh
+            - -ec
+            - |
+              # SPIRE must be able to fork the plugin directly within its container. Copy the plugin into a volume that can be mounted where SPIRE can execute it.
```
faisal-memon, line 136: Maybe move this comment above command:

charts/spire/charts/spire-server/templates/server-resource.yaml
```diff
@@ -121,8 +121,41 @@ spec:
...
+              cp -a /bin/busybox /cel/busybox
+          volumeMounts:
+            - name: cel
+              mountPath: /cel
+          imagePullPolicy: {{ .Values.credentialComposer.cel.image.pullPolicy }}
+        - name: init-cel2
```
faisal-memon, line 142: What does this have to be done twice?

kfox1111: a lot of plugins do not have any shell in them. just the spire plugin. That's better for dealing with CVE's. This plugin is like that.

We need a cp binary that is able to run within the plugin container long enough to copy the plugin to an emptydir so it can be made available in the spire-server container, so spire-server can start it. The first container copies a busybox binary into a place it can be used to copy the plugin into the right place. The second container uses busybox to copy the plugin into the right place. :/

It would be really really nice if plugins could be started in their own container and the spire-server could just talk to them via unix socket... Until spire-server supports that, this is the only way. :(

charts/spire/charts/spire-server/templates/server-resource.yaml
```diff
@@ -121,8 +121,41 @@ spec:
...
       {{- include "spire-lib.default_cluster_priority_class_name" . | nindent 6 }}
-      {{- if or (gt (len .Values.initContainers) 0) (and .Values.upstreamAuthority.certManager.enabled .Values.upstreamAuthority.certManager.ca.create) .Values.nodeAttestor.tpmDirect.enabled $needsChown }}
+      {{- if or (gt (len .Values.initContainers) 0) (and .Values.upstreamAuthority.certManager.enabled .Values.upstreamAuthority.certManager.ca.create) .Values.nodeAttestor.tpmDirect.enabled .Values.credentialComposer.cel.enabled $needsChown }}
       initContainers:
+        {{- if .Values.credentialComposer.cel.enabled }}
+        - name: init-cel
+          securityContext:
+            {{- include "spire-lib.securitycontext" . | nindent 12 }}
+          image: {{ template "spire-lib.image" (dict "appVersion" $.Chart.AppVersion "image" .Values.tools.busybox.image "global" .Values.global) }}
+          # SPIRE must be able to fork the plugin directly within its container. Copy a busybox so that the plugin can be copied into the right place.
```
faisal-memon, line 131:
```suggestion
          # SPIRE must be able to fork the plugin directly within its container. First copy a busybox so that the plugin can be copied into the right place.
```

charts/spire/charts/spire-server/templates/server-resource.yaml
```diff
@@ -121,8 +121,41 @@ spec:
...
+              cp -a /bin/busybox /cel/busybox
+          volumeMounts:
+            - name: cel
+              mountPath: /cel
+          imagePullPolicy: {{ .Values.credentialComposer.cel.image.pullPolicy }}
+        - name: init-cel2
+          securityContext:
+            {{- include "spire-lib.securitycontext" . | nindent 12 }}
+          image: {{ template "spire-lib.image" (dict "appVersion" $.Chart.AppVersion "image" .Values.credentialComposer.cel.image "global" .Values.global) }}
+          # SPIRE must be able to fork the plugin directly within its container. Copy the plugin into a volume that can be mounted where SPIRE can execute it.
```
faisal-memon, line 146:
```suggestion
          # Second, use the previously copied busybox to copy the plugin into a volume that can be mounted where SPIRE can execute it.
```

charts/spire/charts/spire-server/templates/server-resource.yaml
faisal-memon: Wow ok. Thanks for explaining

faisal-memon (approved):

PR #591, opened 2025-05-21
Initial spike support

kfox1111:


v0lkan:
Thanks for the tremendous effort Kevin 🙏 .

I'll check this out and add my comments; hopefully this Fiday.

At a quick glance, it looks good.

charts/spire/charts/spike-keeper/templates/_helpers.tpl
```diff
@@ -0,0 +1,83 @@
...
+  {{- else if and (dig "spire" "recommendations" "enabled" false .Values.global) (dig "spire" "recommendations" "namespaceLayout" true .Values.global) }}
+    {{- if ne (len (dig "spire" "namespaces" "server" "name" "" .Values.global)) 0 }}
+      {{- .Values.global.spire.namespaces.server.name }}
+    {{- else }}
+      {{- printf "spire-server" }}
+    {{- end }}
+  {{- else -}}
+    {{- .Release.Namespace -}}
+  {{- end -}}
+{{- end -}}
```
v0lkan, line 41: Can we default spike-related namespaces to `spike-system` if nothing is provided?

Right now, we install SPIKE to the `default` namespace, by default.

charts/spire/charts/spire-server/templates/controller-manager-cluster-ids.yaml
```diff
@@ -38,8 +53,8 @@ matchLabels:
 {{-   if eq ($root.Values.controllerManager.enabled | toString) "true" }}
 {{-     if or (not (hasKey $value "enabled")) (eq ($value.enabled | toString) "true") }}
 {{-       $type := dig "type" "base" $value }}
-{{-       if not (has $type (list "base" "raw" "child-servers" "oidc-discovery-provider" "test-keys")) }}
-{{-         fail (printf "Type given: %s, must be one of [base, raw, child-servers, oidc-discovery-provider, test-keys]" $type) }}
+{{-       if not (has $type (list "base" "raw" "child-servers" "oidc-discovery-provider" "spike-keeper" "spike-nexus" "spike-pilot" "test-keys")) }}
+{{-         fail (printf "Type given: %s, must be one of [base, raw, child-servers, oidc-discovery-provider, spike-keeper, spike-nexus, spike-pilot, test-keys]" $type) }}
 {{-       end }}
 {{-       $namespaceSelector := deepCopy (dig "namespaceSelector" (dict) $value) }}
```
v0lkan, line 59: namespace selector resolves to something like this by default:

```
  namespaceSelector:
    matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: NotIn
      values:
      - default
      - spire-server
      - spire-system
```

That's rather permissive.

For SPIKE components, it make sense to restrict it to `spike-system` (or whatever namespace the component is overridden to).

Giving SPIKE component an id if it's in spire-system namespace, kind of tingles my spider senses a bit :) .

v0lkan: In my sample sandbox, I used kubernetes-standard annotations (`app.kubernetes.io/name` etc) but if it's not aligned with helm-charts conventions, that's find with me.

pasting my current clusterspiffeid for reference:

```
apiVersion: spire.spiffe.io/v1alpha1
kind: ClusterSPIFFEID
metadata:
  name: spike-keeper
  labels:
    app.kubernetes.io/name: spike-keeper
    app.kubernetes.io/component: keeper
    app.kubernetes.io/part-of: spike
spec:
  className: "spire-mgmt-spire"
  spiffeIDTemplate: "spiffe://spike.ist/spike/keeper/{{ .PodMeta.Name }}"
  namespaceSelector:
    matchLabels:
      kubernetes.io/metadata.name: spike-system
  podSelector:
    matchLabels:
      app.kubernetes.io/name: spike-keeper
      app.kubernetes.io/component: keeper
```

v0lkan: This is especially important for keeper instances as each keeper is pinned to a specific shard; so it will be useful to have some tracking info in the spiffeID.

v0lkan:
For the record; this values file, transformed the chart to more or less aligned with my sandbox:

```yaml
global:
  openshift: false # If running on openshift, set to true
  spire:
    recommendations:
      enabled: true
    namespaces:
      create: true
    clusterName: spike-demo-cluster
    trustDomain: spike.ist
    caSubject:
      country: US
      organization: SPIKE
      commonName: spike.ist

# Enable all SPIKE-related components
spike-keeper:
  enabled: true
  namespaceOverride: spike-system

spike-nexus:
  enabled: true
  namespaceOverride: spike-system

spike-pilot:
  enabled: true
  namespaceOverride: spike-system

spire-server:
  enabled: true

  controllerManager:
    identities:
      clusterSPIFFEIDs:
        spike-keeper:
          type: raw
          enabled: true
          spiffeIDTemplate: spiffe://{{ .TrustDomain }}/spike/keeper/{{ .PodMeta.Name }}
          namespaceSelector:
            matchExpressions:
              - key: kubernetes.io/metadata.name
                operator: In
                values:
                  - spike-system
        spike-nexus:
          type: raw
          enabled: true
          spiffeIDTemplate: spiffe://{{ .TrustDomain }}/spike/nexus/{{ .PodMeta.Name }}
          namespaceSelector:
            matchExpressions:
              - key: kubernetes.io/metadata.name
                operator: In
                values:
                  - spike-system
        spike-pilot:
          type: raw
          enabled: true
          spiffeIDTemplate: spiffe://{{ .TrustDomain }}/spike/pilot/role/superuser
          namespaceSelector:
            matchExpressions:
              - key: kubernetes.io/metadata.name
                operator: In
                values:
                  - spike-system

spire-agent:
  enabled: true

spiffe-csi-driver:
  enabled: true

upstream:
  enabled: false
```

I will test it out and see how it goes on my system.

charts/spire/charts/spire-server/templates/controller-manager-cluster-ids.yaml
kfox1111: > Can we add podmeta.name to spiffeid template for all SPIKE components?
> 
> ```
>   spiffeIDTemplate: "spiffe://{{ .TrustDomain }}/spike/keeper/{{ .PodMeta.Name }}"
> ```

That can be done. In general the spire devs dont like additional metadata put into ids...

Are you planning on supporting checking individual keeper instances, so you know keeper 1 is keeper 1?

If so, something like this does make sense I think.

v0lkan:
update: a mildy-update version of the above yaml worked.

next up: I will try with "zero values yaml, things should work out of the box" version.

I'll keep here posted as usual.

relevant SPIKE branch:
https://github.com/spiffe/spike/tree/feature/helm
(haven't created a PR yet)

charts/spire/charts/spire-server/templates/controller-manager-cluster-ids.yaml
v0lkan: @kfox1111 yep, 

Since keeper is a statefulset the podmeta name will be something like `spiffe-spike-keeper-1` 

That SPIFFEID can be sanitized for additional security, so that in the code, we can do things like:

* Hey keeper 1, the hash of your shard does not match what I have, I don't remmeber sending a new shard to you either, so I will need manual operator override to accept your new shard. Also, this incident is logged for audit bro -- k thx bye.
* Hey keeper 42, what's the matter with you?! -- I know only 3 keepers, and you are not one of them... so no bueno buddy. I'm also audit logging this incident.
* etc, etc.


As long as the number in pod ids match the array indexes in `SPIKE_NEXUS_KEEPER_PEERS` we'd be good.

--

Of course, it's a security-versus-convenience tradeoff.
Because when we assume a "business rule" of array indexes mapping keeper ids,
then if/when we want to add a new keeper, we'll likely need to recompute and redistribute shards—which is not the end of the world, but some additional code to be implemented.

v0lkan (approved):
:taco: .

charts/spire/charts/spike-nexus/templates/service.yaml
```diff
@@ -0,0 +1,20 @@
+{{ $root := . }}
+apiVersion: v1
+kind: Service
+metadata:
+  namespace: {{ include "spike-keeper.namespace" $root }}
```
v0lkan, line 5: @kfox1111 I believe this should b `spike-nexus.namespace`, no?

v0lkan:
will check out and test the recent values (hopefully tonight if I find the energy :) )

I'll keep this thread posted.

charts/spire/charts/spike-nexus/templates/service.yaml
kfox1111: Good catch. Should be fixed.

v0lkan:
change of plans (or introduction of Sev-zeroes :D ) — tonight testing is off the table :D 

I'll update here as soon as I can with my findings.

faisal-memon (approved):

PR #598, opened 2025-05-28
CSI driver: Support setting securityContext

pratiklotia:
The chart for CSI-driver shows that securityContext can be configured but
it doesn't seem that we are using those values when setting the CSI Daemonset template.

Is this intentional or an oversight? I assumed the latter and created this PR.

If that limitation is intentional, we should not make it configurable in the chart
because it is very confusing.

I haven't tested this, if you have suggestions on how one must go about testing, lmk.

### DIFF
--- a/charts/spire/charts/spiffe-csi-driver/values.yaml
+++ b/charts/spire/charts/spiffe-csi-driver/values.yaml
@@ -99,8 +99,8 @@ podSecurityContext: {}
 ## @param securityContext.privileged Flag for specifying privileged mode
 ##
 securityContext:
-  readOnlyRootFilesystem: true
-  privileged: true
+  # readOnlyRootFilesystem: true
+  # privileged: true
   # runAsNonRoot: true
   # runAsUser: 1000
   # capabilities:


pratiklotia:
[Link to chart](https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spiffe-csi-driver/values.yaml#L101-L108)

faisal-memon:
Thanks @pratiklotia for submitting this. It was likely an oversight

charts/spire/charts/spiffe-csi-driver/README.md
```diff
@@ -52,8 +52,7 @@ A Helm chart to install the SPIFFE CSI driver.
...
 | `podSecurityContext`                          | Security context for CSI driver pods                                                                                                                                     | `{}`                                        |
-| `securityContext.readOnlyRootFilesystem`      | Flag for read only root filesystem                                                                                                                                       | `true`                                      |
-| `securityContext.privileged`                  | Flag for specifying privileged mode                                                                                                                                      | `true`                                      |
+| `securityContext`                             | Security context                                                                                                                                    | `true`                                      |
```
faisal-memon, line 55: I dont understand this change here. Why not keep these values but just make sure they are used?

faisal-memon: Thanks for explaining i get it now. Please continue

pratiklotia: Done.

faisal-memon:
@pratiklotia Need to add a section like this to the values.yaml for csi driver: https://github.com/spiffe/helm-charts-hardened/blob/ce9b3737ff714f061592d8455f1ecec294631373/charts/spire/charts/spire-agent/values.yaml#L55-L62

charts/spire/charts/spiffe-csi-driver/README.md
faisal-memon: Thank you.

pratiklotia:
@faisal-memon let me know if this looks good. TY.

charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -99,8 +99,8 @@ podSecurityContext: {}
 ## @param securityContext.privileged Flag for specifying privileged mode
 ##
 securityContext:
-  readOnlyRootFilesystem: true
-  privileged: true
+  # readOnlyRootFilesystem: true
+  # privileged: true
```
kfox1111, line 103: I think we should either replicate the security context in the settings as it was before the change to make it configurable, or we should make it passthrough but set the defaults to what we had set before when we have recommendations set.

before we were doing:
```
            capabilities:
              drop:
                - all
            privileged: true
```

With the change if the user does nothing, they get instead:
```
{}
```

PR #601, opened 2025-05-29
Fix: Spire Agent securityContext issue

pratiklotia:
The upstream Spire Agent [Dockerfile](https://github.com/spiffe/spire/blob/main/Dockerfile#L72) defaults to running as the root user (UID 0). When the spec enforces `runAsNonRoot: true` and/or `runAsUser: 1000` via the `securityContext`, the Spire Agent container will fail with errors like: `container has runAsNonRoot and image will run as root`.

Fix: Explicitly set the `runAsUser: 0` so it is clear that we are running it as root. Also add a warning that, setting `runAsNonRoot` to true will cause errors. Open to better solutions, just looking to avoid confusion.

### DIFF
--- a/charts/spire/charts/spire-agent/values.yaml
+++ b/charts/spire/charts/spire-agent/values.yaml
@@ -58,8 +58,8 @@ securityContext: {}
   #   drop:
   #   - ALL
   # readOnlyRootFilesystem: true
-  # runAsNonRoot: true
-  # runAsUser: 1000
+  # runAsNonRoot: true # WARNING: Setting this to `true` will likely cause the Spire Agent to fail.
+  runAsUser: 0 # Spire Agent typically requires root (UID 0) to function correctly.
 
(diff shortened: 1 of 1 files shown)


kfox1111:
Actually, I think we do run it as non root at times successfully. Looks like even in the recommended config. Its running through:
https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-lib/templates/_helpers.tpl#L297-L308
Then force back to root if tpms are used (so it can reach the tpm):
https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-agent/templates/daemonset.yaml#L24

kfox1111:
Yeah, with recommendations enabled on a non openshift cluster, it has:
```
      securityContext:
        fsGroup: 1000
        fsGroupChangePolicy: OnRootMismatch
        runAsGroup: 1000
        runAsUser: 1000
```

by default.

PR #610, opened 2025-06-17
Add Tornjak manager CRD

Alan-Cha:
Built on top of #597

### DIFF
--- a/charts/spire/charts/spire-server/templates/tornjak-config.yaml
+++ b/charts/spire/charts/spire-server/templates/tornjak-config.yaml
@@ -39,6 +39,13 @@ data:
         }
       }
     {{- end }}
+
+      SPIRECRDManager {
+        plugin_data {
+          classname = "{{ .Values.controllerManager.className }}"
+        }
+      }
+
     {{- if ne .Values.tornjak.config.userManagement.issuer "" }}
       Authenticator "Keycloak" {
         plugin_data {
--- a/charts/spire/charts/spire-server/templates/tornjak-config.yaml
+++ b/charts/spire/charts/spire-server/templates/tornjak-config.yaml
@@ -40,11 +40,13 @@ data:
       }
     {{- end }}
 
+    {{- if .Values.tornjak.config.crdManager.enabled }}
       SPIRECRDManager {
         plugin_data {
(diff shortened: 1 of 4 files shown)


kfox1111:
This seems to have overlap with 597. Is there anything left todo since that has merged?

examples/tornjak/keycloak/README.md
```diff
@@ -49,27 +49,28 @@ The example below demonstrates port forward for local access. In cloud deploymen
...
-> You can set the issuer URL using `--set spire-server.tornjak.config.userManagement.issuer=http://tornjak.tornjak:8080/realms/tornjak`
 >
+> You can set the issuer URL using `--set spire-server.tornjak.config.userManagement.issuer=http://tornjak.tornjak:8080/realms/tornjak`
+
 > [!IMPORTANT]
 > If audience is set, make sure the Tornjak backend `audience` is set correctly. You can set it using:
 > `--set spire-server.tornjak.config.userManagement.audience=your-audience`
->
+
 > [!TIP]
```
Alan-Cha, line 74: These sections were not rendering correctly because there's no separation between them.

Alan-Cha:
@kfox1111 I updated the branch. Should be ready for merging now!

charts/spire/charts/spire-server/templates/tornjak-config.yaml
```diff
@@ -39,13 +39,62 @@ data:
         }
       }
     {{- end }}
+
+      SPIRECRDManager {
+        plugin_data {
+          classname = "{{ .Values.controllerManager.className }}"
+        }
+      }
+
```
kfox1111, line 48: hmm....

Is the crd manager functionality disable-able? If so, is it leaving this section out?

Does it use the permissions of the k8s service account in order to manage the crds?

maia-iyer: Back when it was under development, Tornjak shared the same pod as the controller manager, and spire server (IIRC). Then, it inherited all of the permissions of the controller manager which is what is necessary. Has this changed?

kfox1111:
Closing for now. Please reopen if this is still needed. Thanks!

PR #612, opened 2025-06-23
Update spire-agent and spiffe-csi-driver daemonsets to use node IP from downward API

renanay:
This PR is related to PR https://github.com/spiffe/spire-examples/pull/62 and brings its learning to the charts repository. We currently solve this issue using post rendering but would prefer to have this fixed in the souce code since it impacts OpenShift deployments.

> Fixes issue https://github.com/spiffe/spire/issues/4147 in the spire repo. The spire-agent daemonsets in the examples defaults to using localhost to reach the kubelet API. This fails in some cases and this PR fixes the issue by retrieving the node IP using the downward API and passing it to the spire-agent daemonset configuration.

> The issue is raised in the spire repo but fixes span across spire-examples, spire-tutorials and spire-controller-manager repos under the SPIFFE org.


charts/spire/charts/spiffe-csi-driver/templates/daemonset.yaml
```diff
@@ -89,7 +89,7 @@ spec:
             - name: MY_NODE_NAME
               valueFrom:
                 fieldRef:
-                  fieldPath: spec.nodeName
+                  fieldPath: status.podIP
```
kfox1111, line 92: This pod is not hostNetwork: True, so could cause the node name to float everytime the pod is restarted. It think this will lead to upgrade issues....

kfox1111 (commented):
We have been testing against openshift, and when the openshift flag is set, it relies on nodeName which seems to work.

What is the problem trying to be solved?

charts/spire/charts/spiffe-csi-driver/templates/daemonset.yaml
renanay: I'm not familiar with the code that much. Is there a different solution you would recommend? For the moment this solution is working for us in all openshift environments, not just ocp, but I did miss that this part of the chart is all the time being set, not just for openshift. I can fix it to check if it is an openshift deployment/environment.

kfox1111: no, I mean, the chart seems to already work with on openshift clusters without any changes. We spent considerable time getting it up and running on openshift. I'm curious why the change is needed?

renanay:
> We talked it over and we're thinking an enum based config option change would be best for this...
> 
> `kubeletConnectByHostname` would get renamed to something like `kubeletConnectionType`
> 
> and support `auto`, `hostname`, `hostip`, `localhost`

Cool, thank you! I will start working on the change.

renanay:
I'm closing this PR for now. I don't have the capacity to finish it at this time, sorry.

Issue #613, opened 2025-06-28
Ensure Compliance with Service Name Constraints in Helm Templates


CajuCLC:
**Description:**

The current Helm template for `spire-controller-manager` generates service names that do not comply with the service name constraints outlined in IETF RFC 6335. This results in deployment errors when service names exceed the maximum permitted length of 15 characters. The issue can be observed with the following error:

```
create Pod spire-external-server-0 in StatefulSet spire-external-server failed error: Pod "spire-external-server-0" is invalid: [spec.containers[2].ports[1].name: Invalid value: "prom-cm-ABCDEFG-qa-01": must be no more than 15 characters]
```

The problem originates from the Helm template located at:
[_controller-manager-container.tpl](https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-server/templates/_controller-manager-container.tpl#L96)

To reproduce the issue, the following Helm command was used:

```bash
helm upgrade --install -n spire-mgmt spire spire-nested --repo https://spiffe.github.io/helm-charts-hardened/ \
--set "external-spire-server.kubeConfigs.ABCDEFG-qa-01.kubeConfigBase64=$(cat ../script/ABCDEFG-qa-01.kubeconfig)" \
--set "external-spire-server.kubeConfigs.ABCDEFG-qa-02.kubeConfigBase64=$(cat ../script/ABCDEFG-qa-02.kubeconfig)" \
-f your-values.yaml -f root-values.yaml
```

In this command, the issue arises from the suffix `-ABCDEFG-qa-01`, which is derived from the name specified after `kubeConfigs` in the Helm command arguments, such as `ABCDEFG-qa-01`. This suffix is appended to the service name, resulting in the invalid length.

According to [IETF RFC 6335, Section 5.1](https://www.rfc-archive.org/getrfc.php?rfc=6335), service names:
- Must be no more than 15 characters long.
- Must contain only US-ASCII letters ('A' - 'Z', 'a' - 'z'), digits ('0' - '9'), and hyphens ('-').
- Must contain at least one letter and must not begin or end with a hyphen.
- Hyphens must not be adjacent.

Enforcing these constraints within the Helm template is critical to prevent invalid service name generation and subsequent deployment errors.

CajuCLC:
I added PR #614 , but I wasn't able to test it yet (not home where I can test it).

CajuCLC:
Replying @kfox1111 answer from the PR:

[...]

PR #614, opened 2025-06-28
Ensure Service Name Compliance with RFC 6335

CajuCLC:
Enhances service name compliance in the `spire-controller-manager` Helm template to align with IETF RFC 6335 standards.

### **Changes Include:**
* Introducing a new helper function `valid-service-name` to:
  * Truncate suffixes so that the entire service name does not exceed 15 characters.
  * Replace invalid characters with hyphens using regular expressions to meet character constraints.
  * Remove leading and trailing hyphens, and ensure no consecutive hyphens.
  * Validate that the processed suffix contains at least one letter.
  * Stop template processing with a descriptive error if any of these criteria are not satisfied.

### **Testing:**
* Further testing is required. Initial tests are incomplete, so additional verifications should be conducted to confirm that the template correctly handles various suffix lengths and character types, producing valid service names in different scenarios.

### DIFF
--- a/charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
+++ b/charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
@@ -1,3 +1,23 @@
+{{- define "valid-service-name" -}}
+{{- $name := . -}}
+{{- /* Truncate the name to ensure it fits within the length limit */ -}}
+{{- $maxLength := 15 -}}
+{{- if gt (len $name) $maxLength -}}
+{{- $name = substr $name 0 $maxLength -}}
+{{- end -}}
+{{- /* Replace or sanitize any invalid characters to ensure RFC compliance */ -}}
+{{- $name = regexReplaceAll "[^a-zA-Z0-9-]" $name "-" -}}
+{{- $name = regexReplaceAll "^-+|-+$" $name "" -}}  {{/* Remove leading or trailing hyphens */ -}}
+{{- $name = regexReplaceAll "--+" $name "-" -}}    {{/* Replace multiple hyphens with a single one */ -}}
+{{- if not (regexMatch "[a-zA-Z]" $name) -}}       {{/* Ensure it contains at least one letter */ -}}
+{{- fail "The resulting name must contain at least one letter (A-Z or a-z)." -}}
+{{- end -}}
+{{- if eq (len $name) 0 -}}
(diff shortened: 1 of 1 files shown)


kfox1111:
Thanks for the patch. :)

A lot of good stuff here.

It think there are a few issues with it we need to address...

There are 3 different things here.

### Part 1: ports.name

ports.name has to be RFC 6335, so needs to be <= 15 chars and is pretty strict in its character set.
In addition, it must be unique in the pod. as there are other services in the pod, we need to reserve some space for it. Currently, we prefix 'prom-cm-'. That's 8 chars, so only 7 left.

The truncation in the patch truncates down to 15 not 7, so that truncation is incorrect. But, there is a bigger problem I think. users in this case are also more likely to put the most significant bits on the right hand side. Say the user uses `cluster1` and `cluster2`. Truncating at 7 would just leave `cluster` and `cluster` and we have a conflict. :/. So regular truncation does not feel like the right solution here.

There are a few options I think:
* if too long, drop out ports.name. its not a required property. It does make it a bit harder for a human to debug...
* if too long, hash the name down to 15 chars and ditch the prefix in that case giving as much collision resistance as possible. This does make it a bit hard for a human to use the name.
* right hand side truncation. But a few users may put the most significant bits on the left... So we trade off problems with one type of name for another.

So in all cases, there is no good catch all solution. So we need an escape hatch. We need a way for the end user to set ports.name explicitly when whatever we choose to automatically name it fails.

With manual naming being an option, I'm kind of inclined to go the just drop ports.name when too long route. Its simple, it will always work without conflict, and the user can explicitly set it when they need it, and ignore it if they dont.

### Part 2: containers.name

containers.name isn't restricted to RFC 6335, but DNS hostname. So 63 characters giving us a lot more space. Dashes are allowed. Just not dots. We currently prefix with `spire-controller-manager-` which is 25 characters, leaving 38. That's still a pretty good amount of characters, so I think we can probably set that as an explicit upper limit on kubeconfig names and test for it. If we wanted more, we could probably shorten `spire-controller-manager-` to `scm-` or something.

### Part 3: config file bits

The patch here also changes the naming within the config files. filenames are quite flexible, so I don't think we need to change those bits at all?

CajuCLC:
I agree with all your points. I think we should put this PR on hold and discuss options on the issue instead. What do you think?
Let me add a comment in there.
Issue: #613

PR #623, opened 2025-07-08
spire-server: Replace chown image with busybox

marcofranssen:
This reduces the footprint for this initContainer to just `4.04MB`.

The `cgr.dev/chainguard/bash` image is `35.2MB` in size. Which is like a 31MB image pull saved. The image pull for this container eats up the most time compared to the chown command being run in it.

I have used the same tag as the busybox value to ensure no additional
versions have to be pulled on the node running spire-server components.

Background reason for doing this is to have images that work with AWS ECR pull through.

Alternative I considered is to add more logic and complexity to allow overriding the command (bash -> sh), but that would unnessesary clutter the API, specifically for something as simple as a chown operation.


kfox1111 (commented):
This looks really good! Thanks for working on this.

Could you please update helm-charts-hardened/.github/tests/images.json too so that it will get updates as they come in?

Then, I think this is good to go. :)

PR #626, opened 2025-07-10
Add Agent TTL to Spire Server

CajuCLC:
This pull request introduces a new configuration option, `agentTTL`, to specify the Time-To-Live (TTL) for agent SVIDs in the SPIRE server Helm chart. The changes ensure that this new parameter is documented, configurable, and integrated into the server configuration.

### Addition of `agentTTL` parameter:

* **Documentation update**:
  - Added `agentTTL` to the `README.md` file to describe its purpose and default value (`72h`).

* **Template update**:
  - Integrated `agentTTL` into the `configmap.yaml` template to ensure it is included in the server configuration.

* **Values file update**:
  - Added `agentTTL` to `values.yaml` with a default value of `72h`, allowing users to customize this parameter.


kfox1111 (commented):
Thanks for working on this.

Looking at the server documentation, it mentions that if the setting is unset, it defaults to the value of `default_x509_svid_ttl`. Existing users may have overridden that value and expect that behavior.

On most systems today though, that will be `1h`. Which is significantly less then the proposed `72h`. Its also bigger then the default `caTLL` of `24h`.

So I think for backwards compatibility, we probably want to:
Allow the agentTTL setting to be set to "" and default to that
If blank, don't add the `agent_ttl` option to the config file.

Then the user can still override it with whatever they need and preserve the existing behavior.

kfox1111 (approved):
LGTM. Thanks! :)

PR #627, opened 2025-07-10
Add disk based KeyManager

CajuCLC:
This pull request introduces support for a disk-based Key Manager in the SPIRE agent Helm chart. The changes include updates to configuration files, templates, and documentation to enable and manage this new feature.

### New Feature: Disk-Based Key Manager

* **Documentation Update**: Added a new configuration option, `keyManager.disk.enabled`, to the `README.md` file to document the disk-based Key Manager feature.
* **Configuration Support**: Introduced `keyManager.disk` settings in `values.yaml` to enable the disk-based Key Manager.
* **Template Updates**:
  - Updated `configmap.yaml` to include the disk-based Key Manager configuration under the `plugins` section.
  - Modified `daemonset.yaml` to mount the disk-based Key Manager directory in the container and ensure it is created on the host if it does not exist. [[1]](diffhunk://#diff-e860a3dbff565712efc8045d2a42937c0f58f6994939846799346e253a182e75R259-R263) [[2]](diffhunk://#diff-e860a3dbff565712efc8045d2a42937c0f58f6994939846799346e253a182e75R332-R337)


kfox1111 (commented):
Thanks for the pr.

It looks really good.

The only issue I see with it is that we need to make a configurable for where it lands on disk, and consider the default value of it. Some systems like bootc based ones don't have a writable /opt.

The FHS would lean towards /var/lib/ somewhere.

Also, looking back through values, we already have:
```
persistence:
  type: emptyDir
  hostPath: /var/lib/spire/k8s/agent
```

Is there ever a reason to run keyManager.disk.enabled=true with persistence.type=emtpyDir? Seems like that may break things.

Maybe we make keyManager.disk.enabled depend on persistence.type=hostPath and then just store its stuff under its hostpath.

CajuCLC:
> The only issue I see with it is that we need to make a configurable for where it lands on disk, and consider the default value of it. Some systems like bootc based ones don't have a writable /opt.
> 
> The FHS would lean towards /var/lib/ somewhere.
> 
> Also, looking back through values, we already have:
> 
> ```
> persistence:
>   type: emptyDir
>   hostPath: /var/lib/spire/k8s/agent
> ```

Yeah, shouldn't be a problem to run them in the same path. But the `hostPath` should be changed to `/var/lib/spire` instead?
 
> Is there ever a reason to run keyManager.disk.enabled=true with persistence.type=emtpyDir? Seems like that may break things.
> 
> Maybe we make keyManager.disk.enabled depend on persistence.type=hostPath and then just store its stuff under its hostpath.

I don't know much about Spire to answer the question about disk enabled and type emptydir. For what I have seen it would cause issue since the key would be in the disk, but once the pod restarted, it wouldn't be able to find the SVID. And that would fail.

Let me know if you want me to change the `persistence.hostPath` to `/var/lib/spire` instead. And for the `keymanager.disk` I think it should be also an option of the path you want, maybe you don't want both things in the same folder. The issue with that is mount both, I am not sure it would work.

kfox1111:
Sounds like things will not work right without persistence.type=hostPath. So I do think we just make `keyManager.disk.enabled` dependent on that setting, failing if its not correct, use the existing volume and not make any other setting under the keyManager.disk. We can hardcode the filename within the volume we're managing for the user.

CajuCLC:
> Sounds like things will not work right without persistence.type=hostPath. So I do think we just make `keyManager.disk.enabled` dependent on that setting, failing if its not correct, use the existing volume and not make any other setting under the keyManager.disk. We can hardcode the filename within the volume we're managing for the user.


Sounds good to me. I can make the changes later this week, I am traveling at the moment.

PR #642, opened 2025-08-07
CSI driver: Support setting podSecurityContext and securityContext

alecholmez:
This PR supersedes the below:
https://github.com/spiffe/helm-charts-hardened/pull/598

The `spiffe-csi-driver` chart values file alludes to the security contexts being configurable from within the chart. Previously these values were hardcoded. This PR allows for the override to function correctly along with keeping backwards compatibility with current configuration.


charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -95,12 +95,14 @@ podAnnotations: {}
 podSecurityContext: {}
   # fsGroup: 2000
 
-## @param securityContext.readOnlyRootFilesystem Flag for read only root filesystem
-## @param securityContext.privileged Flag for specifying privileged mode
+## @param securityContext [object] Security context for CSI driver containers
 ##
-securityContext:
+securityContext:
```
kfox1111, line 100:
```suggestion
securityContext:
```

charts/spire/charts/spiffe-csi-driver/templates/daemonset.yaml
```diff
@@ -108,11 +110,7 @@ spec:
               mountPropagation: Bidirectional
               name: mountpoint-dir
           securityContext:
-            readOnlyRootFilesystem: true
```
kfox1111, line 111: This part gets dropped and not replaced...

At a glance, I think the other container may work fine with `readOnlyRootFilesystem: true` as well.

Could you please add it to the values file too and we can test it out?

kfox1111 (commented):
Thanks for working on this. A couple minor issues inline

charts/spire/charts/spiffe-csi-driver/templates/daemonset.yaml
alecholmez: @kfox1111 The `values.yaml` has this as the defaults for the `securityContext`:
```
## @param securityContext [object] Security context for CSI driver containers
##
securityContext: 
  readOnlyRootFilesystem: true
  privileged: true
  capabilities:
    drop:
      - ALL
```
I think that preserves the previous context behavior unless I'm mistaken

alecholmez:
Ping on @kfox1111

PR #646, opened 2025-08-14
Support for Cloud SQL Proxy in GCP

halradaideh:
## Support for Cloud SQL Proxy in GCP

This change adds configuration to enable [Cloud SQL Auth Proxy](https://cloud.google.com/sql/docs/mysql/sql-proxy) support for the `spire-server` Helm chart when running in GCP.  
It leverages **IAM authentication** instead of static credentials, improving security and simplifying secrets management.

**Reference:**  
- [Cloud SQL Auth Proxy Documentation](https://cloud.google.com/sql/docs/mysql/sql-proxy)
- [Blog Example from Google](https://cloud.google.com/blog/products/databases/application-security-with-cloud-sql-iam-database-authentication)

<details>
<summary>Example Usage</summary>

````yaml
spire-server:
  serviceAccount:
    create: true
    annotations:
      iam.gke.io/gcp-service-account: "<SERVICE_ACCOUNT>@<PROJECT_ID>.iam.gserviceaccount.com"
    name: "<SERVICE_ACCOUNT>"
  
  extraContainers:
    - name: cloud-sql-proxy
      image: gcr.io/cloud-sql-connectors/cloud-sql-proxy:2.14.1
      args:
        - "--auto-iam-authn"
        - "--structured-logs"
        - "--port=3306"
        - "<PROJECT_ID>:<REGION>:<DATABASE_INSTANCE_NAME>"

      env:
        - name: GOOGLE_CLOUD_PROJECT
          value: "<PROJECT_ID>"
        - name: GOOGLE_CLOUD_REGION
          value: "<REGION>"

      securityContext:
        runAsNonRoot: true
        runAsUser: 65532
        runAsGroup: 65532
        allowPrivilegeEscalation: false
        capabilities:
          drop: ["ALL"]
        seccompProfile:
          type: RuntimeDefault
  
  dataStore:
    sql:
      databaseType: mysql
      databaseName: spire
      host: 127.0.0.1
      port: 3306
      username: "<SERVICE_ACCOUNT>"
      password: ""
      iamAuth: true
      options: []
      externalSecret:
        enabled: false
        name: ""
        key: ""

### DIFF
--- a/charts/spire/charts/spire-server/values.yaml
+++ b/charts/spire/charts/spire-server/values.yaml
@@ -179,6 +179,9 @@ dataStore:
     ## @param dataStore.sql.clientKeyPath Path to private key for client certificate (MySQL only)
     clientKeyPath: ""
 
+    ## @param dataStore.sql.iamAuth Enable IAM authentication for Cloud SQL (MySQL only). When enabled, password will be excluded from connection string.
+    iamAuth: false
+
     ## When an external source creates the secret. The secret should reside in the same namespace as the spire server
     externalSecret:
       ## @param dataStore.sql.externalSecret.enabled Enable external secret for datastore creds
--- a/charts/spire/charts/spire-server/templates/_helpers.tpl
+++ b/charts/spire/charts/spire-server/templates/_helpers.tpl
@@ -237,18 +237,31 @@ Create the name of the service account to use
 {{- else if or (eq .Values.dataStore.sql.databaseType "mysql") (eq .Values.dataStore.sql.databaseType "aws_mysql") }}
   {{- if eq .Values.dataStore.sql.databaseType "mysql" }}
   {{-   $_ := set $config "database_type" "mysql" }}
-  {{-   $pw = "${DBPW}" }}
-  {{-   $ropw = "${RODBPW}" }}
-  {{- else }}
+  {{-   if .Values.dataStore.sql.iamAuth }}
+  {{-     $pw = "" }}
+  {{-     $ropw = "" }}
+  {{-   else }}
+  {{-     $pw = "${DBPW}" }}
+  {{-     $ropw = "${RODBPW}" }}
+  {{-   end }}
+  {{-   else }}
   {{-   $_ := set $config "database_type" (list (dict "aws_mysql" (dict "region" .Values.dataStore.sql.region))) }}
-  {{- end }}
+  {{-   end }}
   {{- $port := int .Values.dataStore.sql.port | default 3306 }}
   {{- $query := include "spire-server.config-mysql-query" .Values.dataStore.sql.options }}
-  {{- $_ := set $config "connection_string" (printf "%s:%s@tcp(%s:%d)/%s%s" .Values.dataStore.sql.username $pw .Values.dataStore.sql.host $port .Values.dataStore.sql.databaseName $query) }}
+  {{- if .Values.dataStore.sql.iamAuth }}
+  {{-   $_ := set $config "connection_string" (printf "%s@tcp(%s:%d)/%s%s" .Values.dataStore.sql.username .Values.dataStore.sql.host $port .Values.dataStore.sql.databaseName $query) }}
+  {{- else }}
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by kfox1111** [comment]
Interesting. Thanks for working on this.

How does this work? Is the proxy somehow separate from the pods or is there a sidecar container for it?

**Comment by halradaideh** [comment]
It's a sidecar, as you see in the example provided, the proxy is defined as extraContainer 
I have a synchronization problem (race condition) at the moment where the spire server is starting before the proxy is ready

But eventually, it works

The logic in GCP is that you can use a service account with workload identity and IAM policies to grant a Kubernetes service account access to the database
The link provided in the first comment refers to GCP's official documentation

**Comment by halradaideh** [comment]
> > It's a sidecar, as you see in the example provided, the proxy is defined as extraContainer
> 
> Ah. Missed that. Thanks.
> 
> > I have a synchronization problem (race condition) at the moment where the spire server is starting before the proxy is ready
> 
> Should be able to define it as an initContainer instead with
> 
> ```
>    restartPolicy: Always
> ```
> 
> it will make sure it starts first. You may also need a readyness probe on it to guarantee startup ordering.
> 
> Documentation: https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/
> 
> > > But eventually, it works
> > 
> > 
> > The logic in GCP is that you can use a service account with workload identity and IAM policies to grant a Kubernetes service account access to the database The link provided in the first comment refers to GCP's official documentation
> 
> I think we should probably include the sidecar container into the pr/chart too rather then make the user add it to values.

well, my bad, didn't know about it 

PR patched, for your review and approval

File: charts/spire/README.md [hunk 1]
```diff
@@ -206,6 +206,44 @@ helm install -n spire-server spire-crds charts/spire-crds
...
+        - "--auto-iam-authn"
+        - "--port=3306"
+        - "PROJECT_ID:REGION:INSTANCE_NAME"
+
+  dataStore:
+    sql:
+      databaseType: mysql
+      host: 127.0.0.1
+      username: "sa-spire"
+      iamAuth: true
```
kfox1111, line 242: For these sections, can we make the specific service account annotation value,
the adding of the init container,
and the logic behind iamAuth
depend on a different database type?

we already have support for `aws_mysql` and `aws_postgresql`. Maybe we make the values
`gcp_mysql` and `gcp_postgresql` trigger adding the sidecar?

charts/spire/charts/spire-server/values.yaml
```diff
@@ -179,6 +179,8 @@ dataStore:
     ## @param dataStore.sql.clientKeyPath Path to private key for client certificate (MySQL only)
     clientKeyPath: ""
 
+
+
```
kfox1111, line 183: Can you remove the extra white space?

kfox1111 (commented):
Sorry for the delay. Looking good. One minor issue with whitespace

faisal-memon:
Thanks @halradaideh. Pretty interesting way of avoiding the password.

halradaideh:
i've added also the ablity to set load balancer IP
although it's a deprecated flag in Kubernetes, but it's not something that's gonna be removed in the near future
> We (GCP) don't have that alternative yet. So for now you continue using that field until we have it implemented. In any case it will be mentioned in the GKE release notes
> [1] https://github.com/kubernetes/enhancements/pull/1992

reference
https://stackoverflow.com/questions/73750700/what-is-the-replacement-for-the-deprecated-loadbalancerip-attribute-in-services

charts/spire/charts/spiffe-oidc-discovery-provider/templates/service.yaml
```diff
@@ -9,6 +9,9 @@ metadata:
   {{- end }}
 spec:
   type: {{ .Values.service.type }}
+  {{- if and (eq .Values.service.type "LoadBalancer") .Values.service.loadBalancerIP }}
+  loadBalancerIP: {{ .Values.service.loadBalancerIP }}
```
faisal-memon, line 13: Is there any way to have this appear in the readme?

faisal-memon (approved):
Thanks @halradaideh

PR #656, opened 2025-08-28
Add labels to the spiffe-oidc-discovery-provider values.yaml

tuxotron:
When I enabled the OIDC Discovery Provider (K8s), the pod got stuck in the init container (spiffe-helper) because when it requested the x509SVID got rejected because the pod wasn't registered in the system. This was because I have configured my controller manager to only register the pods in my cluster that have the label spiffe.io/spiffe-id: "true".
So when I went to add labels to the spiffe-oidc-discovery-provider chart, the values.yaml file didn't have a reference to the labels, but looking at the template I saw it was just using the labels setting to add labels to the deployment, so I added the setting to the values.yaml file for reference along with a comment.


kfox1111 (approved):
LGTM. Thanks! :)

PR #661, opened 2025-09-11
Spire agent helm chart: allow configuring logFormat

nikotih:
maps to agent configuration log_format ([reference](https://spiffe.io/docs/latest/deploying/spire_agent/#agent-configuration-file))


faisal-memon (approved):
Thanks @nikotih

PR #663, opened 2025-09-11
make spire server's auth_opa_policy_engine configurable in the helm chart

daescha:
adds the capability to configure undefined experimental configs for the spire server and undefined file: contents pairs in the spire server content map. 

The use case for this is I want to change the policy.rego and policy_data.json to adopt a more restrictive policy, but I understand this is a feature SPIRE does not want to encourage users to use. so this PR allows us to use those without forking the helm charts, and without advertising that these are configurable.


daescha:
@kfox1111 I don't follow - as far as I know spire server's experimental.auth_opa_policy_engine.local.policy_data_path and the rego_path ones only take file paths. Heres how I am configuring:

```sh
echo "\
spire-server:
...
  experimental:
    enabled: true
    authOpaPolicyEngine:
      local:
        policy_data_path: /run/spire/config/policy_data.json
        rego_path: /run/spire/config/policy.rego
        use_rego_v1: true
" | helm upgrade --install -n spire-mgmt spire spire/spire -f=- \
  --set-file='spire-server.configMap.additionalMappings.policy_data\.json'=policy_data.json \
  --set-file='spire-server.configMap.additionalMappings.policy\.rego'=policy.rego
```


by the way, I don't really want the other experimental configs, I don't know what they do, is there any way to avoid their effects?

faisal-memon:
> okay @kfox1111, I did as you asked. By the way, is there imminent plans to update to spire 1.13? if so I can remove the use_rego_v1 setting.

Thanks @daescha. You can remove the `use_rego_v1` setting from this PR unless you need it yourself.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -520,4 +530,10 @@ metadata:
 data:
   server.conf: |
     {{- include "spire-lib.reformat-and-yaml2json" (dict "config" (include "spire-server.yaml-config" .) "root" .) | nindent 4 }}
+  {{- if (and .Values.experimental.enabled .Values.experimental.authOpaPolicyEngine.local.policy_data  .Values.experimental.authOpaPolicyEngine.local.rego) }}
```
kfox1111, line 533:
```suggestion
  {{- if (and .Values.experimental.enabled .Values.experimental.authOpaPolicyEngine.local.policy_data .Values.experimental.authOpaPolicyEngine.local.rego) }}
```

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -101,6 +101,16 @@ server:
         - {{ . | quote }}
       {{- end }}
     {{- end }}
+    {{- if (and .authOpaPolicyEngine.local.policy_data  .authOpaPolicyEngine.local.rego) }}
```
kfox1111, line 104:
```suggestion
    {{- if (and .authOpaPolicyEngine.local.policy_data .authOpaPolicyEngine.local.rego) }}
```

kfox1111 (commented):
A couple of nits inline

daescha:
@faisal-memon @kfox1111 okay, I removed the use_rego_v1 key, resolved the nits kevin had with spacing, and removed the additionalMappings key in values.yaml (thought I had removed earlier).

PR #675, opened 2025-10-07
chore: fix gha tests referencing bitnami images

yoctozepto:
Bitnami stopped releasing some images and moved the archival images to bitnamilegacy namespace.

The best fix would be to migrate away from bitnami charts but it would need more engineering.

This change switches to use bitnamilegacy in the meantime.


markgoddard (approved):

kfox1111:
We updated the chart versions and switched the mysql chart to mariadb since bitnami is still supporting that one. So this pr is no longer needed. Thanks for working on in anyway though.

Agreed on longer term needing to switch away from bitnami.

PR #676, opened 2025-10-08
Allow configuring TPM init container security context

jmbass:
## Summary
- keep the TPM plugin and SPIRE server non-root while letting the init copy run as root by default (and remain tunable)
- add values to customize the init container security context, mount path, state directory, and chown behavior to fit strict clusters
- update helpers, configmap references, documentation, and chart metadata to surface the new controls

## Testing
- helm lint charts/spire


jmbass:
Thanks for your response and insights.

You’re right to look at the existing chown init container but I think that unfortunately it doesn’t help in the cases that fail:

It’s only rendered when `$needsChown` is [true](https://github.com/spiffe/helm-charts-hardened/blob/892597576fc64e42b5cebb5d721597240685c5b4/charts/spire/charts/spire-server/templates/server-resource.yaml#L3). `$needsChown` is set when the pod-level `runAsUser` isn’t defined. In our hardened clusters we explicitly set `podSecurityContext.runAsUser: 1000`, so `$needsChown` is false and that init container is not emitted at all :(  . I think moving it earlier wouldn't change that.

Even when it is present, it currently runs `chown -R … /var/lib/spire` [here](https://github.com/spiffe/helm-charts-hardened/blob/892597576fc64e42b5cebb5d721597240685c5b4/charts/spire/charts/spire-server/templates/server-resource.yaml#L186). I understand that this path is used for the legacy datastore.. it never touches `/run/spire/data/tpm-direct`, so the plugin directories would still be owned by root.

Most importantly, the copy step fails before any later init containers run. The TPM init container inherits the pod’s non-root UID, so `cp {{ .Values.nodeAttestor.tpmDirect.pluginPath }} {{ $tpmPluginDst }}` [here](https://github.com/spiffe/helm-charts-hardened/blob/892597576fc64e42b5cebb5d721597240685c5b4/charts/spire/charts/spire-server/templates/server-resource.yaml#L169) hits a permission error on the root-owned PVC before a separate `chown` container could fix anything.
That’s why the patch gives the TPM init container its own `initSecurityContext` (defaults to UID/GID 0) and a `chownAfterCopy` [toggle](https://github.com/spiffe/helm-charts-hardened/pull/676/files#diff-2970cff31c683ad68e57af57ddbb9acf98765a7898efdeff84c8532c260e1dcbR1001). It can write the files, chown them to the pod’s UID, and then the main spire-server container keeps running non-root exactly as before.

PR #691, opened 2025-11-09
use spire-agent.hostCert.resources to set resources for corresponding spire-agent init container

### LINKED ISSUE #690: Cannot set resources for spire-agent init container gather-host-cert, field spire-agent.hostCert.resources is not used
field exists but is not used in spire-agent/daemonset.yaml.

daescha:
closes #690


charts/spire/charts/spire-agent/templates/daemonset.yaml
```diff
@@ -124,6 +124,8 @@ spec:
               echo Mode detected as hostCert.
               {{- end }}
               chmod 644 /hostCert/kubelet.crt
+          resources:
+            {{- toYaml .Values.hostCert.resources | nindent 12 }}
```
kfox1111: I'm wondering if we should just make all the init containers use the same resources setting that the main container uses. since they all run sequentially, that should make for a lot less configurables but keep the resources pretty constant?

daescha: That seems acceptable to me, ill do it.

daescha: I just noticed the spiffe-csi-driver initcontainer set-context also doesnt have resources set. I think Ill do the csi driver here as well then because we are just starting to test our deployments on openshift and have a policy of not allowing any pods without resources set, so we will need that in order to use the csi driver with spire there.

kfox1111 (approved):
Looks good to me! :) @faisal-memon could you please review too?

faisal-memon (approved):

faisal-memon:
Thanks @daescha

PR #701, opened 2025-11-25
Allow configuring spire-agent prometheus listening address

mmckeen:
This allows configuring the prometheus exporter listening address for `spire-agent`.

For example, this can be used to ensure `spire-agent` is listening on `127.0.0.1` when https://github.com/brancz/kube-rbac-proxy is used to protect the prometheus port.


faisal-memon:
Thanks @mmckeen for submitting this.

faisal-memon (approved):

PR #705, opened 2025-12-08
disable hostNetwork on spire-agent daemonset if connect by hostname is true

daescha:
- disables hostNetwork on spire-agent dameonset if kubeletConnectByHostname: true.
- makes dnsPolicy configurable - default is ClusterFirstWithHostNet if kubeletConnectByHostname resolves to false, otherwise, the default kubernetes value.


faisal-memon:
Will this conflict with https://github.com/spiffe/helm-charts-hardened/pull/709?

kfox1111 (approved):
LGTM. Thanks! :)

PR #706, opened 2025-12-08
CSI: Add guard to the validating admission policy for when no volumes exist.

obassett:
Validating Admission Policy to prevent unwanted pods from using the csi driver was generating a lot of errors for my HTTP solver pods with cert-manager as they had no volumes in the spec. This adds a guard clause to check that the spec has a volumes section before trying to validate the contents of it.


kfox1111 (approved):
Good catch. Thanks! :)

PR #721, opened 2026-01-06
Add support for AWS KMS key tagging

hibare:
## Description

This PR adds support for configuring custom tags on AWS KMS keys created by the SPIRE server. Tags are key-value pairs that can be used for resource management, cost allocation, and access control.

Follow up to PR https://github.com/spiffe/spire/pull/6410/

## Changes

- Added `keyManager.awsKMS.keyTags` configuration option in `values.yaml`
- Updated `configmap.yaml` template to render `key_tags` in the AWS KMS plugin configuration
- Added parameter documentation in `README.md`
- Created example configuration in `examples/aws-kms-tags/`

## Related Documentation

- [SPIRE AWS KMS KeyManager Plugin](https://github.com/spiffe/spire/blob/main/doc/plugin_server_keymanager_aws_kms.md#key-tagging)

### DIFF
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -286,9 +286,9 @@ plugins:
         {{- if or (ne .keyPolicy.policy "") (ne .keyPolicy.existingConfigMap "") }}
         key_policy_file: "/run/spire/data/aws-kms-key-policy.json"
         {{- end }}
-        {{- if .keyTags }}
+        {{- with .keyTags }}
         key_tags:
-          {{- range $key, $value := .keyTags }}
+          {{- range $key, $value := . }}
           {{ $key }}: {{ $value | quote }}
           {{- end }}
         {{- end }}


charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -286,6 +286,12 @@ plugins:
         {{- if or (ne .keyPolicy.policy "") (ne .keyPolicy.existingConfigMap "") }}
         key_policy_file: "/run/spire/data/aws-kms-key-policy.json"
         {{- end }}
+        {{- if .keyTags }}
+        key_tags:
+          {{- range $key, $value := .keyTags }}
```
marcofranssen: Nit…

Using `with` we do the `if` condition and change context so we don't have to repeat the values key when using it. Especially useful if navigating longer paths. E.g. `.Values.some.deeply.nested.value`. I think it makes the template slightly cleaner.

```suggestion
        {{- with .keyTags }}
        key_tags:
          {{- range $key, $value := . }}
```

PR #787, opened 2026-04-09
Fix agent.conf generation logic for SVIDStore

dszakallas:
Fix the casing of "svidstore" to "svidStore" in the configuration merging logic for `agent.conf`.

It previously created invalid configuration for SVIDStore by falling through to the `else` branch of the condition and generating PascalCased `SvidStore` thus  causing the `spire-agent` to be stuck in CrashLoopBackOff.

```
spire-agent time="2026-04-09T09:40:31.607601422Z" level=error msg="Agent crashed" error="unsupported plugin type \"SvidStore\""
```

This corrects the logic so the correct `SVIDStore` key is geneated in `agent.conf`.

### DIFF
--- a/charts/spire-lib/templates/_helpers.tpl
+++ b/charts/spire-lib/templates/_helpers.tpl
@@ -202,7 +202,7 @@ Take a copy of the config and merge in .Values.customPlugins and .Values.unsuppo
 {{- $pluginsToMerge := dict "plugins" dict }}
 {{- range $type, $val := .root.Values.customPlugins }}
 {{-   if . }}
-{{-     if eq $type "svidstore" }}
+{{-     if eq $type "svidStore" }}
 {{-       $_ := set $pluginsToMerge.plugins "SVIDStore" (deepCopy $val) }}
 {{-     else }}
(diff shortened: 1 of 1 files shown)


kfox1111:
This looks like a good change, thank you! but we just broke out the spire-lib chart to a different location. Could you please update the pr to target the new location?

dszakallas:
@kfox1111 I rebased on main

kfox1111:
Parsing through the code, it looks like it used to work, but with the wrong name? This means people could be using it successfully with the wrong name? Pretty unlikely but possible. We may want to consider an upgrade note about it. @faisal-memon  what do you think?

dszakallas:
I was trying to use the `"svidstore"` as a key under `spire-agent.customPlugins`, but that fails this assertion:
https://github.com/dszakallas/spire-helm-charts-hardened/blob/fix-svidstore-config/charts/spire/charts/spire-agent/templates/configmap.yaml#L5-L7

kfox1111:
Ah, ok. Thanks for verifying. So it never worked, so we can just fix it without needing to tell anyone how to migrate, as there is nothing to migrate.

kfox1111 (approved):
LGTM. Thanks! :)

PR #798, opened 2026-04-23
allow unsupported built-in key manager plugins to be used

daescha:
currently the field `spire-server.unsupportedBuiltInPlugins.keyManager` exists but all it does is cause the chart templating to fail if it is set and another keymanager is enabled. I would like the ability to actually use its values.

Specifically, I want the ability to use the gcp_kms or azure_key_vaults plugins. I imagine I will make a PR later to add actual support for those. But we need this change on a quicker timeline than what it would take to do that properly and give enough of a buffer for the chart to have a new release so we can use the official version, so I would prefer to get this in and revisit it later.

### DIFF
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -322,7 +322,11 @@ plugins:
   {{- end }}
   {{- end }}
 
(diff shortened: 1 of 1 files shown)


kfox1111 (approved):
LGTM. Thanks! :)

PR #799, opened 2026-04-24
Flatten SPIRE subcharts to root charts

faisal-memon:
## Summary

This PR flattens the remaining SPIRE component charts out of `charts/spire/charts/` into root-level charts under `charts/`, continuing the move toward a clearer umbrella-chart plus component-chart layout.

## What Changed

- moved SPIRE component charts to root-level `charts/*`
- updated `charts/spire` and `charts/spire-nested` to depend on the new root-level chart locations
- added direct `spire-lib` dependencies to moved component charts that render shared library templates
- generated `Chart.lock` files for the moved component charts and refreshed umbrella locks
- expanded local dependency preparation so `make chart-deps` prepares the umbrella charts and moved components consistently
- updated repo docs and helper scripts that referenced the old nested chart paths

## Why

The repo has been in a hybrid state where `charts/spire` was both an umbrella chart and the canonical home for many component charts. This change makes the root-level charts the source of truth and simplifies the next steps toward a cleaner multi-umbrella layout.

`make chart-deps` remains the accepted source-checkout preparation step for umbrella chart installs from a fresh clone.

## Validation

- `make chart-deps`
- `helm template smoke charts/spire`
- `helm template smoke charts/spire-nested`
- `helm lint charts/spire`
- `helm lint charts/spire-nested`
- `helm lint charts/spiffe-csi-driver`
- `helm lint charts/spiffe-oidc-discovery-provider`
- `helm lint charts/spike-keeper`
- `helm lint charts/spike-nexus`
- `helm lint charts/spike-pilot`
- `helm lint charts/spire-agent`
- `helm lint charts/spire-server`
- `helm lint charts/tornjak-frontend`
- `cd tests && GOCACHE=/tmp/go-build-cache GOMODCACHE=/tmp/go-mod-cache go test -count=1 ./...`
- `./helm-docs.sh`

## Notes

- local `make lint` progressed through chart-testing and then stopped because `yamllint` was not installed in the local environment
- this PR is intentionally focused on repo layout and dependency wiring rather than behavioral chart changes


kfox1111:
This is looking good. Thanks for working on this :)

3 issues I think.

1. the chart testing is failing because it generally assumes a chart is self sufficient. This isnt true of most of the charts. an agent wont work without a server. We can skip testing those individually and test them through testing the aggregate chart still. So I think we just need to exclude them.
2. ~~For the broken out charts here, we may want to mark their versions the same as the spire chart they came from to indicate they are further along then 0.1.0. Maybe we do that in a different PR though.~~
3. we need to tweak the image updater and the image list https://github.com/spiffe/helm-charts-hardened/blob/main/.github/tests/images.json to support the values files in the right place. (It still would be very cool to break this tool out to its own github action too)

kfox1111:
I was thinking about it a bit more. if we dont cut a release of the subcharts yet, we can merge the pr with 0.1.0 versions. We can bump the other versions in another pr before releasing them individually.

PR #816, opened 2026-05-02
Implement easy Bottom Turtle HA support in the charts

kfox1111:


charts/spire-nested/values.yaml
```diff
@@ -401,13 +405,350 @@ external-spire-server:
...
+    spire:
+      ## @param internal-spire-server-bottom-turtle-ha-b.upstreamAuthority.spire.enabled Enable upstream SPIRE server
+      enabled: true
+      ## @param internal-spire-server-bottom-turtle-ha-b.upstreamAuthority.spire.upstreamDriver Use an upstream driver for authentication
+      upstreamDriver: upstream-b.csi.spiffe.io
+      server:
+        ## @param internal-spire-server-bottom-turtle-ha-b.upstreamAuthority.spire.server.nameOverride The name override setting of the root SPIRE server
+        nameOverride: root-server
+        ## @param internal-spire-server-bottom-turtle-ha-b.upstreamAuthority.spire.server.address Address for upstream Spire server
+        address: "spire-server-a"
```
faisal-memon: should this be spire server b?

charts/spire-nested/values.yaml
```diff
@@ -401,13 +405,350 @@ external-spire-server:
...
+      mode: emptyDir
+  healthChecks:
+    ## @param downstream-spire-agent-bottom-turtle-ha-b.healthChecks.port Health check port
+    port: 9982
+  telemetry:
+    prometheus:
+      ## @param downstream-spire-agent-bottom-turtle-ha-b.telemetry.prometheus.port Prometheus port to use
+      port: 9990
+  ## @param downstream-spire-agent-bottom-turtle-ha-b.socketPath Socket path to use
+  socketPath: /var/run/spire/agent/sockets/a/csi.spiffe.io/public/spire-agent.sock
```
faisal-memon: should this be `sockets/b`?

faisal-memon:
I like the work here overall. I found a couple of minor copy/paste issues, but the overall direction makes sense.

A few things I’d consider for future iterations, not required for merge:

- Create a separate chart for this architecture instead of continuing to grow spire-nested.
- There’s a lot of duplication between the A and B paths. That may not be fully reducible given the static Helm dependency structure, but it would be nice to centralize more of the A/B-specific values to reduce drift and copy/paste mistakes.

kfox1111:
> I like the work here overall. I found a couple of minor copy/paste issues, but the overall direction makes sense.

Thanks. Fixed the issues.

> A few things I’d consider for future iterations, not required for merge:
> 
> * Create a separate chart for this architecture instead of continuing to grow spire-nested.
> * There’s a lot of duplication between the A and B paths. That may not be fully reducible given the static Helm dependency structure, but it would be nice to centralize more of the A/B-specific values to reduce drift and copy/paste mistakes.

I figured it belonged in the nested chart as it is a spire nesting. just that the spire servers are being contacted via the hosts services. But, yeah, there is a case for making it its own chart.

That would add more work to the release process though. But maybe thats ok if we automate more of the release process like whats happening with version number generation.

What do you think? Should we stick with the spire-nested layout since it fits the nesting definition, or break it out to its own chart? I have a very slight preference to leave it in spire-nesting, due to ease, but just a slight preference. I could go either way.

Yeah, I'd like to see more sharing in values too. But not sure how to do that short of generating the values.yaml files with code, which was pushed back hard against before. I'm open to it though.

faisal-memon (approved):

PR #819, opened 2026-05-04
fix: workaround for SPIRE Controller Manager log config disrespect

jeengbe:
Upstream bug that CM only respects logger configuration from CLI, not fixed yet: https://github.com/spiffe/spire-controller-manager/pull/667.

Without patch:

image

With patch:

image


kfox1111:
https://github.com/spiffe/helm-charts-hardened/pull/840 should have fixed this.

PR #825, opened 2026-05-12
aws node attester: add org verification support

### LINKED ISSUE #824: spire-server chart: expose AWS IID verify_organization (org membership validation)
The SPIRE aws_iid server node attestor can validate that a node’s AWS account is in your Organization via `verify_organization `(management account ID, role in that account for organizations:ListAccounts, optional region and cache TTL). See [docs](https://github.com/spiffe/spire/blob/v1.14.5//doc/plugin_server_nodeattestor_aws_iid.md#enabling-aws-node-attestation-organization-validation).

The `verify_organization` is not configurable through chart values currently. We should fix this to enable full support for the aws plugin.

pratiklotia:
Fixes #824


charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -246,6 +253,18 @@ plugins:
         {{- if ne .assumeRole "" }}
         assume_role: {{ .assumeRole | quote }}
         {{- end }}
+        {{- if and .verifyOrganization (eq (.verifyOrganization.enabled | toString) "true") }}
+        {{- $vo := .verifyOrganization }}
```
kfox1111:
```suggestion
        {{- with .verifyOrganization }}
```
this will make '.' be defined as .verifyOrganziation within the with block.

you can then get rid of the $vo's in the block. like, .managementAccountId should work
and add an extra {{- end }} at the end of the block to match with the with.

kfox1111 (commented):
This is looking really good. thanks for working on this.

One minor suggestion inline

kfox1111 (approved):
LGTM. Thanks! :)

PR #826, opened 2026-05-12
wire ratelimit configuration option in spire-server configMap

heybronson:
wires `ratelimit` configuration options into the spire-server configMap.

https://spiffe.io/docs/latest/deploying/spire_server/#server-configuration-file


charts/spire/charts/spire-server/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: spire-server
 description: A Helm chart to install the SPIRE server.
 type: application
-version: 0.1.0
```
kfox1111, line 5: please revert the version bumps. we will bump the versions before release all at once. thanks! :)

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -120,6 +120,11 @@ server:
     {{- end }}
   {{- end }}
   {{- end }}
+  {{- with .Values.ratelimit }}
```
kfox1111: Can we update it to rateLimit to follow our naming convention?

kfox1111 (commented):
This looks pretty good. some little things inline.

One consideration is that k8s often hides the actual ips of incoming requests. so this may be less useful then desired. Just FYI.

charts/spire/charts/spire-server/templates/configmap.yaml
heybronson: done!

charts/spire/charts/spire-server/Chart.yaml
heybronson: reverted!

kfox1111 (approved):
LGTM. Thanks! :)

heybronson:
I'm not able to merge :/

PR #829, opened 2026-05-15
Configure jwt_issuer in SPIRE OIDC Provider

g-gaston:
The SPIRE OIDC Discovery Provider binary supports a top-level `jwt_issuer` configuration key. When set, the provider returns that exact string as the `issuer` field in the OIDC discovery document (`.well-known/openid-configuration`) regardless of how the request was routed. When unset, it derives `issuer` from the inbound HTTP Host header. This chart did not render that key into the OIDC provider config, leaving the discovery document Host-derived even when the operator had a fixed issuer in mind.

## Why this matters

OpenID Connect Discovery requires the discovery doc's `issuer` to be byte-equal to the JWT `iss` claim. Conformant verifiers reject the chain when the two differ. Production OIDC consumers routinely reach the discovery endpoint at a URL different from the canonical issuer:

- a load balancer, ingress, or NodePort exposes the provider on an IP or host different from the canonical issuer name;
- TLS terminates at a hostname different from the one advertised to clients;
- the discovery URL is fetched by an internal service (e.g. the API server in a private cluster) over a different DNS view than external clients use;
- a pinned issuer URL is contractually required and must survive infrastructure changes that move the actual service endpoint.

In all of these the JWT's `iss` claim is a logical, stable URL; the discovery doc must report that same value, or downstream verifiers reject the tokens.

## Current chart behavior

The chart already has `global.spire.jwtIssuer` (and a subchart-local `jwtIssuer`), resolved by the `spire-lib.jwt-issuer` helper to `global.spire.jwtIssuer` -> subchart-local `jwtIssuer` -> `https://oidc-discovery.<trustDomain>`. The spire-server subchart writes that helper's result unconditionally as `jwt_issuer:` into the server's config -- this controls the `iss` claim of every JWT-SVID the server mints. In the OIDC subchart, however, the same helper was only used for two things:

1. as the strict-mode assertion gate (fails the render when the resolved value is the `example.org` default);
2. as the default source for `config.jwtDomain` (the Host allow-list).

It was never written into the rendered OIDC provider configuration file. The asymmetry means the chart shipped a structurally invalid OIDC setup by default: the spire-server signs JWTs with `iss = <resolved issuer>`, while the OIDC discovery endpoint advertises whatever Host header was used to reach it. The only way to correct that today is to patch the rendered ConfigMap out of band (`kubectl patch`, a CMP, a kustomize post-renderer), which defeats the purpose of the chart.

## Backward compatibility and behavior changes

Operators who set `global.spire.jwtIssuer` will see one additional `jwt_issuer:` line in the rendered OIDC ConfigMap. The discovery doc's `issuer` will start returning that pinned value instead of being Host-derived, bringing the chain into spec compliance; this is a fix for any spec-compliant verifier that previously rejected tokens. Operators who only set the OIDC subchart-local `jwtIssuer` see the same fix applied via the helper's fallback chain. Operators with nothing set will see the new line default to `https://oidc-discovery.<trustDomain>`, matching what the spire-server config already emits today.


kfox1111:
Sorry for the delay on this, and thanks for working on it. I've been out sick, and am trying to finish up a big pr. We're going to need to have a conversation around the best approach for this as well. It works against some of the things we decided upon for behavior, so need to carefully consider the ramifications of the proposed changes.

g-gaston:
> Sorry for the delay on this, and thanks for working on it. I've been out sick, and am trying to finish up a big pr. We're going to need to have a conversation around the best approach for this as well. It works against some of the things we decided upon for behavior, so need to carefully consider the ramifications of the proposed changes.

@kfox1111 no worries! What are those decisions? Happy to change the approach if it fits those better (also glad to go read if there are written somewhere). Hope you feel better!

kfox1111:
@marcofranssen Has using the internal to k8s routes ever been useful outside of the tests we do? If they really do need to validate the jwtIssuer in the token against it, I don't think those routes ever worked properly? If they never worked, there are more options to fix this issue.

g-gaston:
@kfox1111 @marcofranssen 
gentle bump 🙂

kfox1111:
I had a chat with Marco. He's going to look at doing a bit of testing to see if the other code path is dead.

kfox1111:
We havent found a case so far where just forcing the issuer to always be set to whatever it is set to doesnt make sense.

So, maybe we go down that route? Then we dont need an extra variable to set.

g-gaston:
> We havent found a case so far where just forcing the issuer to always be set to whatever it is set to doesnt make sense.
> 
> So, maybe we go down that route? Then we dont need an extra variable to set.

@kfox1111 sounds good to me, even simpler of a change. I'll update the code

g-gaston:
@kfox1111 done 🙂

kfox1111:
One last thing and I think this is good to go.

in charts/spire/README.md in the upgrade section, could you add something like:

0.30.X
* The OIDC discovery issuer is now set automatically. We do not anticipate any negative impact; however, please verify your OIDC provider’s integration with other services during your upgrade testing.

g-gaston:
> One last thing and I think this is good to go.
> 
> in charts/spire/README.md in the upgrade section, could you add something like:
> 
> 0.30.X
> 
> * The OIDC discovery issuer is now set automatically. We do not anticipate any negative impact; however, please verify your OIDC provider’s integration with other services during your upgrade testing.

Done!

kfox1111 (approved):
LGTM. Thanks! :)

g-gaston:
@kfox1111 when could I expect a new version to be cut containing this fix? Thanks!

kfox1111:
@g-gaston I'm working on putting the finishing touches on https://github.com/spiffe/helm-charts-hardened/pull/860 then once merged, I think we're ready for another release.

PR #831, opened 2026-05-17
Automate release bumps from the current branch

faisal-memon:
## Summary
- add a --from-current-branch mode to release-chart.sh
- make current-branch dry runs stop before commit, push, or PR creation
- reuse the current dependent-chart helper interface and reduce repeated Helm repo refreshes
- handle missing prior chart tags when building release notes

## Testing
- bash -n release-chart.sh
- ./release-chart.sh --chart spire-lib --bump patch --from-current-branch --dry-run


kfox1111 (approved):
These sorts of changes can be pretty hard to test without actually working through the release to see what happens.

It looks good to me. Lets find out. :)

Thanks!

PR #835, opened 2026-05-21
feat(spire-server): add terminationGracePeriodSeconds

heybronson:
Adds two new top-level values to the spire-server chart:

- `terminationGracePeriodSeconds` (nullable, pod-spec level): overrides the default 30s termination grace period. Useful when the server is behind a load balancer that needs time to deregister the target.

- `lifecycle` (object, container level): lifecycle hooks for the spire-server container. The primary use case is a preStop hook to hold the pod alive while the load balancer deregisters the target before SIGTERM is sent:


kfox1111:
terminationGracePeriodSeconds makes sense to me.

lifecycle feels a little more problematic. Shouldn't the chart provide any lifecycle hooks needed for proper functioning itself? What should be in there that isn't already there?

heybronson:
terminationGracePeriodSeconds and lifecycle are complementary.

The primary use case for lifecycle is to add a delay before the container is stopped or terminated, giving load balancers enough time to deregister the target. The reason the entire lifecycle block is generally exposed is that users may configure it differently depending on their specific load balancer configurations.

kfox1111:
> The primary use case for lifecycle is to add a delay before the container is stopped or terminated, giving load balancers enough time to deregister the target. The reason the entire lifecycle block is generally exposed is that users may configure it differently depending on their specific load balancer configurations.

I guess I'm saying, we may not want users putting arbitrary lifecycle hooks in there, which could interfere with upgrading the chart over time.

If all that is needed is a user configurable timeout, maybe we add that directly to the chart?

faisal-memon:
@heybronson thanks for opening this pr. since `terminationGracePeriodSeconds` we have agreement one can we split that off to a seperate pr while we disuss lifecycle?

kfox1111 (approved):
LGTM. Thanks! :)

PR #842, opened 2026-06-06
Fix release workflow path trigger

faisal-memon:
## Summary
- fix the release workflow path filter to point at helm-release.yaml
- ensure workflow-only changes to the release workflow actually trigger on pushes to release

## Testing
- inspected failing release workflow behavior after rerun
- verified the path filter was pointing at helm-release.yml while the actual workflow file is helm-release.yaml


kfox1111 (approved):
LGTM. Thanks! :)

PR #856, opened 2026-06-23
feat(spire-server): support x509pop external_pki caBundlePath

sylvain-blot-sp:
Render ca_bundle_path when nodeAttestor.x509POP.mode is external_pki and caBundlePath is set. Keeps existing spiffe-exchange rendering for spiffe mode.


sylvain-blot-sp (commented):
all good

anas-zurkiyeh-sp (approved):

charts/spire/charts/spire-server/values.yaml
```diff
@@ -1126,8 +1126,10 @@ nodeAttestor:
   x509POP:
     ## @param nodeAttestor.x509POP.enabled Enable the x509_popg node attestor
     enabled: false
-    ## @param nodeAttestor.x509POP.mode What mode to set the plugin to. Currently only spiffe mode is supported
+    ## @param nodeAttestor.x509POP.mode Plugin mode: spiffe (exchange) or external_pki (enrollment CA bundle)
     mode: spiffe
+    ## @param nodeAttestor.x509POP.caBundlePath Path to PEM CA bundle on disk when mode is external_pki
+    caBundlePath: ""
```
kfox1111: In other cases, we have not just made it a path in the container, but provided config to point at the configmap, and also an option to provide the contents of the configmap.

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -261,6 +261,11 @@ plugins:
     x509pop:
       plugin_data:
         mode: {{ .mode }}
+        {{- if and (eq .mode "external_pki") .caBundlePath }}
```
kfox1111: could mode be changed to externalPKI to be more consistent with our naming conventions?

kfox1111 (commented):
Thanks for working on this!

Some things inline.

charts/spire/charts/spire-server/README.md
```diff
@@ -457,7 +457,11 @@ In order to run Tornjak with simple HTTP Connection only, make sure you don't cr
...
+| `nodeAttestor.x509POP.caBundle.existingConfigMap`                                            | Name of a ConfigMap containing a `ca-bundle.pem` key with the PEM CA bundle.                                                                                                                                                                                                                                                                                                                         | `""`                                                                                                        |
+| `nodeAttestor.x509POP.caBundlePath`                                                          | Path to PEM CA bundle on disk when mode is externalPKI. Overrides chart-managed caBundle mount when set.                                                                                                                                                                                                                                                                                             | `""`                                                                                                        |
```
kfox1111: This all looks good I think, except I'm not sure we want to support arbitrary paths in the container. There has been resistance to letting the user poke in arbitrary things in arbitrary places in the container. It makes it harder to guarantee upgrades will work smoothly. Are there any identified issues with the configmap interface that the path would be used to work around? If not, lets remove the path value for now.

kfox1111:
The templates were a lot easier to read/maintain in the previous version before adding all the ()'s and guards.

I spent the last few days digging and it seems like the unit test framework is kind of broken. Its somehow not loading the values files from the chart, so everything is undefined. I poked around a bit and couldnt figure out why this is so far.

So we have two options. Clutter up the templates with guards that only affect the unit tests, or skip the unit tests for now until we can figure out how to fix the unit testing framework.

I'm inclined to recommend the latter.

savitha-qs:
I'm interested in this feature. Is it going to make the next release?

kfox1111:
Depends on if someone has time to work on it before then.

savitha-qs:
@sylvain-blot-sp are you planning on continuing this work? If not, I can help with this PR or create a new one. Please let me know, thanks!

savitha-qs:
@kfox1111 I've opened https://github.com/spiffe/helm-charts-hardened/pull/908. Please take a look. Thank you!

savitha-qs:
@sylvain-blot-sp FYI - this PR can be closed. https://github.com/spiffe/helm-charts-hardened/pull/908 (which is a continuation of your work) has been merged.

PR #859, opened 2026-06-24
Enable easy plugin loading

### LINKED ISSUE #589: plugin copying settings
It is hard to load in plugins as they need to be copied into the spire-server/spire-agent containers. They don't always have a shell either, so need a helper to load them. This can be made generic, so its easy to give a list of containers / a binary in the container that are plugins to load and do the heavy lifting for the end user.

kfox1111:
customPluings have to be loaded into the main container somehow. Extend the existing cel plugin loader to allow users to easily specify an image to load it from.

Before this feature, you had to do this:
```
spire-server:
  initContainers:
    - name: install-custom-plugin-busybox
      image: busybox:1.37.0-uclibc
      command: ["busybox", "sh", "-ec", "cp -a /bin/busybox /custom-plugin/busybox"]
      volumeMounts:
        - name: plugin-staging
          mountPath: /custom-plugin
    - name: install-custom-plugin
      image: ghcr.io/spiffe/spire-credentialcomposer-identity-exchange:dev
      imagePullPolicy: Never
      command: ["/custom-plugin/busybox", "sh", "-ec", "/custom-plugin/busybox cp -a /ko-app/spire-credentialcomposer-identity-exchange /custom-plugin && /custom-plugin/busybox rm /custom-plugin/busybox"]
      volumeMounts:
        - name: plugin-staging
          mountPath: /custom-plugin
  extraVolumes:
    - name: plugin-staging
      emptyDir: {}
  extraVolumeMounts:
    - name: plugin-staging
      mountPath: /custom-plugin/spire-credentialcomposer-identity-exchange
      subPath: spire-credentialcomposer-identity-exchange
  customPlugins:
    credentialComposer:
      spire-credentialcomposer-identity-exchange:
        plugin_cmd: /custom-plugin/spire-credentialcomposer-identity-exchange
        plugin_checksum: ""
        plugin_data: {}
```

With this feature, you can now do:
```
spire-server:
  customPlugins:
    credentialComposer:
      spire-credentialcomposer-identity-exchange:
        image:
          registry: ghcr.io
          repository: spiffe/spire-credentialcomposer-identity-exchange
          tag: dev
          pullPolicy: Never
        plugin_cmd: /ko-app/spire-credentialcomposer-identity-exchange
        plugin_checksum: ""
        plugin_data: {}
```

Fixes: https://github.com/spiffe/helm-charts-hardened/issues/589


faisal-memon:
I think there is one behavioral edge case worth fixing before merge.

The new image-based plugin loader rewrites every imaged `plugin_cmd` to `/plugins/<basename>` in `charts/spire-lib/templates/_helpers.tpl`, and then the init-container logic copies each plugin binary into that shared directory by basename in the server/agent templates.

That means if two custom plugins come from different images but both use the same filename, for example `/bin/plugin`, they will both land at `/plugins/plugin` and the later copy will overwrite the earlier one. SPIRE will then execute the wrong binary for at least one configured plugin.

It would be safer to preserve a unique destination per plugin instance instead of flattening to `base(plugin_cmd)`.

faisal-memon (approved):

PR #860, opened 2026-06-26
Experimental support for spire-identity-exchange

kfox1111:


faisal-memon:
Non-blocking follow-up: the generated README still shows `AppVersion: 0.1.0` in the badge, while `charts/spire-identity-exchange/Chart.yaml` now has `appVersion: "v0.2.0"`. It looks like the docs may just need one more regeneration pass so the published chart metadata and README stay in sync.

faisal-memon:
Non-blocking follow-up: the comment under `tls:` in `charts/spire-identity-exchange/values.yaml` still says "If none are enabled, connections wont be TLS encrypted," but the template now requires exactly one TLS source to be enabled and fails otherwise. It would be good to update that text so the values docs match the current behavior.

faisal-memon (approved):

PR #864, opened 2026-07-01
add config to allow disabling jwt svids

### LINKED ISSUE #863: add config to set `disable_jwt_svids` on spire server
dont use jwts. want to disable the creation of kms keys for them.

daescha:
closes #863


charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -57,7 +57,7 @@ server:
   {{- end }}
   jwt_issuer: {{ include "spire-lib.jwt-issuer" . | quote }}
   audit_log_enabled: {{ .Values.auditLogEnabled }}
-
+  disable_jwt_svids: {{.Values.disableJwtSvids}}
```
kfox1111: Please add spaces around the value.

charts/spire/charts/spire-server/README.md
```diff
@@ -165,6 +165,7 @@ In order to run Tornjak with simple HTTP Connection only, make sure you don't cr
...
 | `clusterDomain`                                                                              | This is the value of your clusters `kubeadm init --service-dns-domain` flag                                                                                                                                                                                                                                                                                                                          | `cluster.local`                                                                                             |
+| `disableJwtSvids`                                                                            | If true, completely disables JWT-SVID functionality. The server will not generate JWT keys, sign JWT-SVIDs, or implement JWT-related API calls. This is useful for deployments that don’t need JWT-SVIDs support.                                                                                                                                                                                    | `false`                                                                                                     |
```
kfox1111: I had a quick chat with the other maintainers

We have a convention of requiring acronyms to be capitalized when not the first word. We also would like to avoid two acronyms together, and also prefer enable settings rather then disable settings. So, how about we change the config option to:
jwtSVIDSupport
Defaulting true.

kfox1111 (commented):
Thanks for working on this. Some things inline.

charts/spire/charts/spire-server/README.md
daescha: was wondering about this, thanks for the review. Resolved.

kfox1111 (approved):
LGTM. Thanks! :)

PR #866, opened 2026-07-02
Add default labels to child-servers clusterSPIFFEIDs for external-root-spire-server-full

### LINKED ISSUE #851: SPIRE controller manager memory usage scales with child cluster workload count in nested SPIRE setup
### Versions
  - Helm chart: 0.27.0
  - SPIRE server image: 1.13.0
  
  ### Environment

  Root cluster node pool
  - 7 × c7g.4xlarge (16 vCPU, 32 GiB)
  - 3 × c7g.2xlarge (8 vCPU, 16 GiB)

  ---
  ### Description 
In a nested SPIRE setup, the SPIRE controller manager (spiffe/spire-controller-manager) on the external (root) server syncs all ClusterSPIFFEIDs from its managing child cluster, rather than only the entries it requires. This over-syncing causes memory consumption to scale with the total workload count on the child cluster, which is not expected behaviour for operators running nested SPIRE at scale.

  ### Observations
With ~4,000 workload pods on the child cluster, the SPIRE controller manager on the external server requires at least 512 MiB of memory to operate stably. The memory required increases proportionally to the number of workloads deployed on the child cluster.

  ### Expectations
The SPIRE controller manager should only reconcile the ClusterSPIFFEIDs it requires, independent of the total workload count on the child cluster. This would make the memory usage fixed rather than scale with workload size.

###   Why raising the memory limit is not a safe workaround
The helm chart propagates external-spire-server controller manager settings to all dynamically created controller manager instances — one per child cluster. Raising the memory limit therefore multiplies the resource allocation across every child cluster's controller manager. On a root cluster with finite node capacity, this leads to resource contention and can cause SPIRE server scheduling failures.
  
###   Proposed fix
Introduce a label selector on ClusterSPIFFEIDs so the SPIRE controller manager can filter which entries it reconciles, scoping it to only the entries it requires.

The label selector must be applied at the Kubernetes API level, so that not all ClusterSPIFFEIDs are loaded in memory.

### Important 
1. This fix would require a coordinated change: ClusterSPIFFEIDs must first be labelled (e.g. via an updated helm chart template) before deploying a controller manager that filters by label, to avoid a window where the controller manager reconciles no entries and breaks attestation.

JoelGoh92:
Part of solution to close https://github.com/spiffe/helm-charts-hardened/issues/851

Introduces a default label `spire.spiffe.io/child-server: "true"` on the `child-servers` ClusterSPIFFEID generated by the `nestedChildFull` chart setup.

This is a prerequisite for a follow-up controller manager fix that will restrict the informer cache on the root cluster's external controller manager to only the ClusterSPIFFEIDs it actually needs (i.e. the child SPIRE server identity), rather than caching all workload ClusterSPIFFEIDs from the child cluster. 

This is the root cause of the memory scaling issue reported in #851.


kfox1111 (approved):
LGTM. Thanks! :)

PR #868, opened 2026-07-04
Introduce chart `spire-tcp-broker`

matheuscscp:
SPIRE PR: https://github.com/spiffe/spire/pull/6915

Blocked by: https://github.com/spiffe/spire/issues/7116

Introduce:

The `spire-tcp-broker` top-level chart allows either a single Deployment in the same namespace where the core SPIRE components are deployed i.e. where the top-level `spire` chart is deployed, or one Deployment per Broker e.g. a dedicated Helm release can be deployed to the `flux-system` namespace and it will serve all and only the Flux controllers. For every Helm release, the resulting ServiceAccount must be allow-listed in the `spire-server` subchart value `nodeAttestor.k8sPSAT.podAgentsCluster.serviceAccountAllowList`.


kfox1111:
Thanks for starting work on this very important set of functionality.

This has some overlap with https://github.com/spiffe/helm-charts-hardened/pull/860 which should merge soon, and would benefit from the x509pop changes from that pr. Can this be rebased on top of that pr?

Also, I'm  wondering if i would be better to break the pr into two parts? I think the spire-agent changes to the existing daemonset will be easier to get working/merged. It might be less cluttered if we review the tcp part in a followon pr? There is a couple ways we could go with it.

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -359,6 +359,13 @@ socketAlternate:
     pullPolicy: IfNotPresent
     tag: latest@sha256:90041f375e30f41aa7e0390075d8a69dc61900771d52fa98d63ee5d03d866a58
 
+broker:
+  ## @param broker.enabled Enable the experimental SPIFFE Broker API over Unix domain socket only
+  enabled: false
+  ## @param broker.brokers [array] Broker SPIFFE IDs allowed to use the Broker API
+  brokers: []
+  # - id: spiffe://example.org/istio-ztunnel
```
kfox1111: Not a fully baked idea.... 

Helm doesnt deal as well with list merging as it does with dict merging. There are other places where we may want to enable brokers.
spire-ha-agent for example. This will be used in the nested chart.... It would be nice if the user could enable istio-ztunnel without having to also specify the spire-ha-agent again.

Maybe there is a way to rework this so that it is a dict instead of a list.

charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -81,18 +99,31 @@ agent:
...
+    broker:
+      socket_path: {{ include "spire-agent.broker-socket-path" . | quote }}
+      brokers:
+        {{- range .Values.broker.brokers }}
+        - id: {{ .id | quote }}
+          allowed_reference_types:
+            - type_url: {{ include "spire-agent.broker-workload-pid-reference-type" $ | quote }}
+              allow_over_tcp: false
+            - type_url: {{ include "spire-agent.broker-kubernetes-object-reference-type" $ | quote }}
+              allow_over_tcp: false
```
kfox1111: This is starting to crystallize in my head...
What if we made it in values:

```
broker:
  istio-ztunnel:
    enabled: false
    idTemplate: spiffe://{{ .TrustDomain }}/istio-ztunnel
    allowed_reference_types:
      - type_url: type.googleapis.com/spiffe.broker.KubernetesObjectReference
         allow_over_tcp: false
  spire-ha-agent:
    enabled: false
    idTemplate: spiffe://{{ .TrustDomain }}/spire-ha-agent
    allowed_reference_types:
      - type_url: type.googleapis.com/spiffe.broker.WorkloadPIDReference
        allow_over_tcp: false
```

then the user can just --set broker.istio-ztunnel.enabled=true

broker api would be enabled if any entry is enabled.

charts/spire/charts/spire-agent/templates/_helpers.tpl
```diff
@@ -126,6 +126,10 @@ Create the name of the service account to use
 {{- print .Values.socketPath }}
 {{- end }}
 
+{{- define "spire-agent.broker-socket-path" -}}
+/run/spire/broker-sockets/broker.sock
```
kfox1111: This cant be hardcoded... There are k8s distros where this path wont work.

We currently have this structure for the admin/delegated api:
```
sockets:
  hostBasePath: /run/spire/agent/sockets
  admin:
    enabled: false
    mountOnHost: false
```

Can we fit it into that model?

kfox1111 (commented):
Just going to review the spire-agent related changes, not the tcp stuff for now.

kfox1111:
> Also, I'm wondering if i would be better to break the pr into two parts? I think the spire-agent changes to the existing daemonset will be easier to get working/merged. It might be less cluttered if we review the tcp part in a followon pr? There is a couple ways we could go with it.

Actually, until there is a spire release with the broker in it, we cant ship any of the pr, so breaking it up doesnt matter unless that happens first.

kfox1111:
For testing in the pr, you can override the image tags to use nightly. We can remove the tags before merging.

charts/spire/charts/spire-broker-over-tcp/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+name: spire-broker-over-tcp
```
kfox1111: There are two hard problems in computer science.
Naming things, cache invalidation, and off by one errors.

Naming things can take a while... we should probably start the conversation now.

Should we simplify this to just spire-broker? spire-broker-over-tcp does kind of make sense but also is kind of verbose/technical. What about spire-network-broker?

Also, I'm still thinking we may want multiple physical or logical brokers. so maybe we use the plural.
spire-network-brokers?

kfox1111 (changes requested):

charts/spire/charts/spire-server/templates/configmap.yaml
```diff
@@ -214,6 +193,16 @@ plugins:
...
           {{- end }}
+          {{- if eq (.brokerOverTCP.enabled | toString) "true" }}
+        - {{ include "spire-lib.broker-over-tcp-cluster-name" $root }}:
+            service_account_allow_list: {{ include "spire-server.brokerOverTCPServiceAccountAllowedList" $root | trim }}
+            use_pod_uid_for_agent_id: true
+            audience: {{ .audience }}
+            allowed_node_label_keys:
+              {{ toYaml .allowedNodeLabelKeys | nindent 14 }}
+            allowed_pod_label_keys:
+              {{ toYaml .allowedPodLabelKeys | nindent 14 }}
```
kfox1111: I'm not sure the same allowedNodeLabelKeys and allowedPodLabelKeys should be shared between the daemonset and tcp deployments.

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -181,7 +181,24 @@ Create the name of the service account to use
 {{-   end }}
 {{-   $list | toJson }}
 {{- else }}
-[{{ printf "%s:%s-agent" $releaseNamespace .Release.Name | quote }}]
+{{-   list (printf "%s:%s-agent" $releaseNamespace .Release.Name) | toJson }}
+{{- end }}
+{{- end }}
+
+{{- define "spire-server.brokerOverTCPServiceAccountAllowedList" }}
+{{- $releaseNamespace := include "spire-server.agent-namespace" . }}
```
kfox1111: I think this should be "spire-server.namespace" rather then agent-namespace. the spire-system namespace is intended for things that need privilege (like, hostPath mounts). The broker deployment shouldnt need privilege and should be locked down in the spire-server namespace.

charts/spire/charts/spire-broker-over-tcp/Chart.yaml
kfox1111: Also, we're resisting adding new charts under charts/spire/charts. We're breaking them out as we go. It should be added under charts, and then add a dependency in charts/spire/Charts.yaml please.

matheuscscp: Yeah I think this chart is better off as a top-level one that users can deploy multiple times, e.g. I can deploy this chart in the `flux-system` namespace and have NetworkPolicy only allow traffic into it from the namespace.

charts/spire/charts/spire-agent/templates/configmap.yaml
matheuscscp: I really don't think hard-coding other CNCF projects here is a good idea. Other projects (including Istio competitors) can implement the same integration. Having one dict for each such project is simply not smart.

charts/spire/charts/spire-agent/values.yaml
matheuscscp: I don't think we need to support merging here, do we?

matheuscscp: I have removed the istio use case from this PR and will open a separate PR.

charts/spire/charts/spire-agent/templates/configmap.yaml
matheuscscp: I have removed the istio use case from this PR and will open a separate PR.

charts/spire/charts/spire-agent/templates/_helpers.tpl
matheuscscp: I have removed the istio use case from this PR and will open a separate PR.

charts/spire/charts/spire-broker-over-tcp/Chart.yaml
matheuscscp: Fixed, thanks!

charts/spire/charts/spire-server/templates/configmap.yaml
matheuscscp: Fixed, thanks!

charts/spire/charts/spire-server/templates/_helpers.tpl
matheuscscp: Fixed, thanks!

charts/spire-tcp-broker/templates/agent-alias.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: spire.spiffe.io/v1alpha1
+kind: ClusterStaticEntry
+metadata:
+  name: {{ printf "%s-agent-alias" (include "spire-tcp-broker.cluster-fullname" .) | quote }}
+  labels:
+    {{- include "spire-tcp-broker.labels" . | nindent 4 }}
+spec:
+  className: {{ .Values.controllerManagerClassName | quote }}
```
kfox1111, line 8: controllerManager feels wrong.... but not entirely sure why.... The CR's have their own class name to be tagged with... spiffeClassName?

charts/spire-tcp-broker/templates/configmap.yaml
```diff
@@ -0,0 +1,193 @@
...
+          ]
+        },
+        {{- end }}
+      ]
+    }
+  }
+}
+
+plugins {
+  NodeAttestor "k8s_psat" {
```
kfox1111, line 136: There is a decision point here...

I will need x509pop support. It works in upstream 1.15.2 today. So if supported in this PR, we could merge the PR with a workable solution before 1.15.3. you could implement it here, or I could implement a pr on top of this one we merge together before merging this pr.

The other option is to wait until after 1.15.3, merge without x509pop support, and I'll add it on as a followon pr., shortly thereafter. This will likely take longer, wallclock time. Which way would you like to go?

charts/spire-tcp-broker/templates/configmap.yaml
```diff
@@ -0,0 +1,193 @@
...
+    plugin_data {
+      cluster = {{ .Values.clusterName | quote }}
```
kfox1111, line 138: We should consider making clusterName macro and spiffeClassName  set of globals and helper functions in spire-lib if we are starting to use them across charts. That way its more likely a user can reuse the settings across charts.

charts/spire-tcp-broker/templates/configmap.yaml
```diff
@@ -0,0 +1,193 @@
...
+
+{{- define "spire-tcp-broker.hcl-config" -}}
+{{- $trustDomain := include "spire-lib.trust-domain" . }}
+agent {
+  data_dir = "/var/lib/spire"
+  log_level = {{ .Values.logLevel | quote }}
+  log_format = {{ .Values.logFormat | quote }}
+  rebootstrap_mode = {{ .Values.rebootstrapMode | quote }}
+  rebootstrap_delay = {{ .Values.rebootstrapDelay | quote }}
+  server_address = {{ .Values.server.address | quote }}
```
kfox1111, line 103: something like:
server_address: {{ include "spire-tcp-broker.server-address" . | quote }}
or maybe we break it out over to spire-lib.

charts/spire-tcp-broker/templates/configmap.yaml
```diff
@@ -0,0 +1,193 @@
...
+    plugin_data {
+      cluster = {{ .Values.clusterName | quote }}
+    }
+  }
+
+  KeyManager "memory" {
+    plugin_data {}
+  }
+
+  WorkloadAttestor "k8s" {
```
kfox1111, line 146: Discussion point. Not sure how to feel here yet....

The spire-identity-exchange will not have this plugin enabled.
flux will want it enabled.
Should we have a worlkloadAttestor.k8s.enabled, defaulted off, with it settable to true? Then those that want this particular config can ask for it.

I guess there is some crystal ball'ing here on how it will be used in the future beyond those two use cases too.

charts/spire-tcp-broker/templates/deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+          securityContext:
+            {{- include "spire-lib.securitycontext" . | nindent 12 }}
+          env:
+            - name: PATH
+              value: "/opt/spire/bin:/bin"
+            {{- with .Values.extraEnvVars }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          ports:
+            - containerPort: {{ .Values.service.port }}
```
kfox1111, line 53: This should use a local hardcoded port IMO.
The service should be made to use lower port numbers, preferably 443, but will not be allowed to be used by the pod when running as non root. The pod I think has no reason to run as root so would have to use a higher port like 8443

charts/spire-tcp-broker/templates/deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+          projected:
+            sources:
+            - serviceAccountToken:
+                path: spire-agent
+                expirationSeconds: 7200
+                audience: spire-server
+        {{- if eq (len .Values.trustBundleURL) 0 }}
+        - name: spire-bundle
+          configMap:
+            name: {{ include "spire-lib.bundle-configmap" . | quote }}
```
kfox1111, line 116: This is a bit problematic...

In a production recommended setup, the trust bundle configmap here is only in the spire-system namespace, where privileged pods run.

This pod though can and should be highly restricted so it should should by default be in the spire-server namespace.

We may need a new config option to make the trust bundle available in the spire-server namespace too.

charts/spire-tcp-broker/values.yaml
```diff
@@ -0,0 +1,152 @@
...
+  ## @param readinessProbe.initialDelaySeconds Initial delay seconds for probe
+  initialDelaySeconds: 10
+  ## @param readinessProbe.periodSeconds Period seconds for probe
+  periodSeconds: 30
+
+service:
+  ## @param service.type Service type
+  type: ClusterIP
+  ## @param service.port Service port for the Broker API
+  port: 8443
```
kfox1111, line 113:
```suggestion
  port: 443
```

charts/spire-tcp-broker/values.yaml
```diff
@@ -0,0 +1,152 @@
...
+  ## @param service.annotations [object] Annotations for service resource
+  annotations: {}
+
+## @param controllerManagerClassName SPIRE Controller Manager class name that reconciles the agent alias
+controllerManagerClassName: ""
+## @param workloadAgentAlias SPIFFE ID of the agent alias that parents the Broker ClusterStaticEntries
+workloadAgentAlias: ""
+
+
```
kfox1111, line 122:
```suggestion
```

charts/spire-tcp-broker/values.yaml
```diff
@@ -0,0 +1,152 @@
...
+## @param brokers [array] Named Brokers allowed to use the Broker API over TCP. namespace defaults to the release namespace and serviceAccountName defaults to the Broker name.
+brokers: []
+# - name: broker
+#   namespace: broker-namespace # defaults to the release namespace
+#   serviceAccountName: broker-service-account # defaults to the Broker name
```
kfox1111, line 127: This assumes that the broker will always be a k8s service account and generates the spiffe id from it. I dont think this is always true. So I think we should make this just be the spiffe id instead.

charts/spire/charts/spire-server/README.md
```diff
@@ -446,6 +446,10 @@ In order to run Tornjak with simple HTTP Connection only, make sure you don't cr
...
+| `nodeAttestor.k8sPSAT.podAgentsCluster.clusterName`                                                           | Name of an additional k8s_psat cluster for pod-based SPIRE agents. Empty disables the cluster.                                                                                                                                                                                                                                                                                                       | `""`                                                                                                        |
```
kfox1111, line 449: Naming things is hard...

Not sure what Cluster in podAgentsCluster means. Maybe we should just call the section podAgents?

tests/integration/spire-tcp-broker/client/main.go
```diff
@@ -0,0 +1,216 @@
+// brokerclient is a manual integration-test driver for the SPIFFE Broker API.
```
kfox1111, line 1: This is a great tool, but feels like the wrong place for this. Shouldn't this be part of the spire repo? or maybe spire-examples? We will want to use this tool to test in other pipelines.

tests/integration/spire-tcp-broker/run-tests.sh
```diff
@@ -0,0 +1,407 @@
...
+
+set -Eeuo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../../.." && pwd)"
+
+KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-chart-testing}"
+KUBE_CONTEXT="kind-${KIND_CLUSTER_NAME}"
+K8S="${K8S:-}"
+SPIRE_SERVER_IMAGE="${SPIRE_SERVER_IMAGE:-ghcr.io/spiffe/spire-server:1.15.1}"
```
kfox1111: Having the version in a variable name here means we will need to bump it with every release which will be painful. These variables should function as an override if set, otherwise, fall back to whats in the chart.

tests/integration/spire-tcp-broker/run-tests.sh
```diff
@@ -0,0 +1,407 @@
...
+  --set-string "spire-server.image.tag=$(image_tag "${SPIRE_SERVER_IMAGE}")" \
+  --set-string spire-server.image.pullPolicy=IfNotPresent \
+  --set-string "spire-agent.image.registry=$(image_registry "${SPIRE_AGENT_IMAGE}")" \
+  --set-string "spire-agent.image.repository=$(image_repository "${SPIRE_AGENT_IMAGE}")" \
+  --set-string "spire-agent.image.tag=$(image_tag "${SPIRE_AGENT_IMAGE}")" \
+  --set-string spire-agent.image.pullPolicy=IfNotPresent \
+  --wait \
+  --timeout 5m
+
+mapfile -t agent_nodes < <(
```
kfox1111, line 239: I had to look this up. You taught me a new trick. Love it. Thanks! :)

tests/integration/spire-tcp-broker/run-tests.sh
```diff
@@ -0,0 +1,407 @@
...
+        .metadata.managedFields,
+        .metadata.resourceVersion,
+        .metadata.uid
+      ) |
+      .metadata.namespace = $namespace
+    ' |
+    k apply -f -
+}
+
+copy_bundle "${APP_A_NAMESPACE}"
```
kfox1111, line 289: Hacky.... We need a way to configure the server to create the bundles in the right place, or better would be for the agents to be able to pull the trust bundle directly from the configmap in a different namespace.

tests/integration/spire-tcp-broker/run-tests.sh
```diff
@@ -0,0 +1,407 @@
...
+
+install_broker() {
+  local namespace="$1"
+  local values="$2"
+  local workload_agent_alias="$3"
+
+  h upgrade --install "${BROKER_RELEASE}" "${REPO_ROOT}/charts/spire-tcp-broker" \
+    --namespace "${namespace}" \
+    --values "${values}" \
+    --set-string clusterName=broker-pods \
```
kfox1111, line 300: Shouldnt clusterName match what the spire chart was installed with?

tests/integration/spire-tcp-broker/run-tests.sh
```diff
@@ -0,0 +1,407 @@
...
+    --set-string controllerManagerClassName=spire-server-spire \
+    --set-string server.address=spire-server.spire-server \
+    --set-string "workloadAgentAlias=${workload_agent_alias}" \
```
kfox1111, line 303: Wait.... this is the AGENT_ALIAS_A below, and that includes the node uid in it? How can a normal user deploy in this way? This feels very wrong.

There has to be a better way...

Couldnt this just be automatically set based on the release name and have it be unique enough?

charts/spire-tcp-broker/templates/agent-alias.yaml
```diff
@@ -0,0 +1,14 @@
...
+  spiffeID: {{ include "spire-tcp-broker.agent-alias-id" . | quote }}
+  parentID: {{ include "spire-tcp-broker.server-id" . | quote }}
```
kfox1111, line 10: I had to look parentID up, only to have it just be:
```
spiffe://{{ include "spire-lib.trust-domain" . }}/spire/server
```

and only ever used once in the pr. We should just collapse it to that value here to simplify the code.

tests/integration/spire-tcp-broker/run-tests.sh
```diff
@@ -0,0 +1,407 @@
...
+  local namespace="$1"
+  local name="$2"
+  local service_account="$3"
+  local node="$4"
+  shift 4
+
+  echo "=== Test case: ${namespace}/${name} ==="
+  k delete job "${name}" --namespace "${namespace}" --ignore-not-found --wait=true >/dev/null
+
+  jq -n \
```
kfox1111, line 356: Just to share, but not needing to fix,
You could also do this with a bash heredoc. Bash can substitute the variables directly then without an extra set of mapping

Ex:
```
PERSON=You
cat <<EOF | kubectl apply -f -
#Hello $PERSON
#pod spec bits here
EOF

File: charts/spire/charts/spire-server/templates/configmap.yaml [hunk 26]
```diff
@@ -214,6 +224,16 @@ plugins:
             allowed_pod_label_keys:
               {{ toYaml .allowedPodLabelKeys | nindent 14 }}
           {{- end }}
+          {{- if ne .podAgentsCluster.clusterName "" }}
```
kfox1111, line 227: ...

So we have to create another named "cluster" for the podAgents. :/

Feels a little hacky...

What if we remove the .clusterName requirement and just make the following line:
```
- {{ include "spire-lib.cluster-name" $root }}-broker:
```

Then it can be autoset?

User can specify still if they want, but dont have to.

charts/spire-tcp-broker/templates/agent-alias.yaml
```diff
@@ -0,0 +1,14 @@
...
+  selectors:
+    - {{ printf "k8s_psat:cluster:%s" .Values.clusterName | quote }}
```
kfox1111, line 12: Can we make the clusterName optional? I think it should be unique enough without it. Then it should be possible for the user to skip configuring clusterName if they want.

kfox1111 (commented):
Initial review

charts/spire-tcp-broker/templates/agent-alias.yaml
matheuscscp: Isn't the className a global setting in the spire-controller-manager?

charts/spire-tcp-broker/templates/configmap.yaml
matheuscscp: Isn't spire-identity-exchange a different chart? How will it use this one?

charts/spire-tcp-broker/templates/deployment.yaml
matheuscscp: What if I want to deploy this chart in the namespace of the broker application? Would that be possible?

charts/spire-tcp-broker/values.yaml
matheuscscp: I'm confused: in the other thread you said to use 8443

charts/spire-tcp-broker/templates/configmap.yaml
kfox1111: Yes.

Its currently embedding the spire-agent as a sidecar in its own deployment. Ideally, it would use the spire-tcp-broker to scale the agents separately from the spire-identity-exchange pods.

charts/spire-tcp-broker/templates/deployment.yaml
kfox1111: Should be possible. But currently installing it anywhere but the spire-system namespace wont have the trust bundle configmap.

charts/spire-tcp-broker/values.yaml
kfox1111: service.port = 443,  port in the pod, 8443.

charts/spire-tcp-broker/templates/deployment.yaml
matheuscscp: Hmm I think this is not a concern of the chart, but of how you get stuff into your Kubernetes cluster. Via GitOps I could for example use this: https://fluxoperator.dev/docs/crd/resourceset/#copying-data-from-existing-configmaps-and-secrets So I can copy the ConfigMap from spire-system to the broker namespace

tests/integration/spire-tcp-broker/run-tests.sh
kfox1111: And found out today that it doesn't work on macos out of the box because bash is told for mapfile. :/

PR #876, opened 2026-07-13
SPIRE Agent support for Broker API

kfox1111:


charts/spire/charts/spire-agent/templates/configmap.yaml
```diff
@@ -81,9 +81,26 @@ agent:
...
+      {{- if .Values.brokerAPI.tcp.enabled }}
+      bind_address: {{ .Values.brokerAPI.tcp.bindAddress | quote }}
+      {{- end }}
+      brokers:
+        {{- range .Values.brokerAPI.brokers }}
+        - id: {{ tpl .idTemplate (dict "trustDomain" $trustDomain) }}
+          {{- with .allowedReferenceTypes }}
+          allowed_reference_types:
+            - type_url: {{ .typeURL | quote }}
+              allow_over_tcp:A {{ .allowOverTCP }}
```
matheuscscp:
```suggestion
              allow_over_tcp: {{ .allowOverTCP }}
```

charts/spire/charts/spire-agent/values.yaml
```diff
@@ -467,10 +467,43 @@ sockets:
...
   hostPath: /var/lib/spire/k8s/agent
+
+brokerAPI:
+  tcp:
+    ## @param brokerAPI.tcp.enabled Enable the broker api endpoint
+    enabled: false
+    ## @param brokerAPI.tcp.bindAddress The tcp address to bind to
+    bindAddress: 0.0.0.0:8788
+  brokers:
+    istio-ztunnel:
```
matheuscscp: Why would you put a downstream consumer in this chart?

kfox1111: Hmm.... That is a good question....

On the one hand, we want to make it as easy as possible to use for common cases. On the other, it may become unmaintainable if there is to many things that keep getting added here...

I can see a few different policies that could apply, and maybe apply to controller manager identities as well:
* No brokers/identities are ever preconfigured, even off by default. (pure rule)
* Every broker/identity will always be accepted but may be off by default (pure rule)

Every other option is a compromise between those two.

brokers/identities are preconfigured, even off by default IF:
* there is support for the component in the chart (like, spike)
* the component is part of the spiffe umbrella but not directly otherwise supported in the chart
* the component is part of the cncf umbrella but not directly otherwise supported in the chart. the argument being, they are sister projects so get special consideration.

The last option would allow preconfigured setups to easily enable
* flux
* istio
* cilium

matheuscscp:
Hey @kfox1111 this PR is missing things I'm adding to #868. I'd like to respectfully request you to please step aside and let me finish implementing the feature end-to-end.

kfox1111:
> Hey @kfox1111 this PR is missing things I'm adding to #868. I'd like to respectfully request you to please step aside and let me finish implementing the feature end-to-end.

I dont understand. #868 is implementing the tcp broker with a Deployment. Which is great. This PR is just turning on broker support in the existing daemonset.

The two PR's are orthogonal, and both are important. Did I miss something?

The helm charts hardened project has a strong preference for several smaller PR's rather then one gigantic one to ease review, and merge smaller bits of functionality as its ready. Progress can be made on each of them without blocking the other.

matheuscscp:
Yep, you asked me to split the PR and I did the first step yesterday, removing the DaemonSet part from there

matheuscscp:
I didn't have time to file one for the DaemonSet since then.

matheuscscp:
> The helm charts hardened project has a strong preference for several smaller PR's rather then one gigantic one to ease review, and merge smaller bits of functionality as its ready.

Looks like in #868 I'm still ~half the budget set by #860

kfox1111:
Then why is two people with limited time working together worse then one doing it all the lifting themselves?

matheuscscp:
I don't think we're on a rush here, and I've been pushing through Broker API all the way in my free time because it's fun and I'd like to finish the job, but a maintainer is snatching away the fun I'm having as an opensource contributor here. I'm losing my motivation to contribute to this project... If you don't understand that, I don't know how to explain it better.

kfox1111:
@matheuscscp  and I chatted on slack. We will go forward with this PR for this limited subset of the work so the overall broker work doesn't slow down spire-ha-agent & istio ztunnel support.

charts/spire/charts/spire-agent/values.yaml
kfox1111: Since this is contentious, lets table the conversation in this PR. We can always add entries later in a followon PR, if we decide to include such functionality.

matheuscscp: It's fine to add things like `spire-ha-agent` that are defined in this repository, as they are very core, you can change their integration point atomically here. But it makes no sense to add istio things here, just like it makes no sense to add flux things in the other PR. Thanks for removing.

faisal-memon (approved):

PR #884, opened 2026-07-17
Add broker suport to the spire-ha-agent

kfox1111:


faisal-memon:
I think there is still one blocking issue in the new Broker API config rendering for the SPIRE agent.

In `charts/spire/charts/spire-agent/templates/configmap.yaml`, the template now emits `brokers` entries with extra wrapper levels, effectively shaped like:

```yaml
brokers:
  - broker:
    - id: ...
      allowed_reference_types:
        - allowed_reference_types:
            type_url: ...
            allow_over_tcp: ...
```

But SPIRE expects each `brokers` item to be a broker object with `id` and `allowed_reference_types` directly on that object, and each allowed reference type item to contain `type_url` / `allow_over_tcp` directly. The extra `broker:` and nested `allowed_reference_types:` wrapper levels change the schema, so I would expect broker-enabled installs to render an agent config that SPIRE will reject or ignore.

kfox1111:
That issue is with the hcl1 parser being a little bit weird with parsing yaml. Its tested and works.

faisal-memon (approved):

PR #889, opened 2026-07-22
Add externalSecret support to spire-server kubeConfigs

sabsari:
## What

Lets each `spire-server` `kubeConfigs` entry reference an externally-managed
Kubernetes Secret instead of embedding a plaintext/base64 kubeconfig in values.
Entries may reference different Secrets and mix inline with external ones.

## Details

- Each `kubeConfigs.<name>` entry sets exactly one of `kubeConfig`,
  `kubeConfigBase64`, or the new `externalSecret`:
  ```yaml
  kubeConfigs:
    cluster-a:
      kubeConfig: |            # existing inline
        ...
    cluster-b:
      externalSecret:
        name: my-secret        # required
        key: cluster-b         # optional, defaults to the entry name
  ```
- The `kubeconfigs` volume becomes projected: inline entries come from the
  chart-generated Secret, each `externalSecret` entry from its own Secret.
  Consumer mount paths (`/kubeconfigs/<entry>`) are unchanged.
- The generated Secret holds only inline entries, and is skipped when all
  entries are external.

## Notes

- Backward compatible: the volume type changes from `secret` to `projected`, but
  mounted paths and contents are identical (one rolling restart on upgrade).

### DIFF
--- a/tests/unit/spire_test.go
+++ b/tests/unit/spire_test.go
@@ -257,4 +257,35 @@ spiffe-oidc-discovery-provider:
 			Expect(serverCM).Should(ContainSubstring(`"jwt_issuer": "https://oidc-discovery.example.org"`))
 		})
 	})
+	Describe("spire-server.kubeConfigs", func() {
+		secretTmpl := "spire/charts/spire-server/templates/kubeconfig-secret.yaml"
+		serverTmpl := "spire/charts/spire-server/templates/server-resource.yaml"
+		It("inline entry generates a Secret and a projected volume source referencing it", func() {
+			objs, err := ValueStringRender(chart, `
+spire-server:
+  kubeConfigs:
+    clustera:
+      kubeConfig: |
+        apiVersion: v1
+        kind: Config
+`)
+			Expect(err).Should(Succeed())
+			Expect(objs[secretTmpl]).Should(ContainSubstring("kind: Secret"))
+			Expect(objs[serverTmpl]).Should(ContainSubstring("projected:"))
+			Expect(objs[serverTmpl]).Should(ContainSubstring("path: clustera"))
+		})
+		It("externalSecret entry wires a projected source and skips the generated Secret", func() {
+			objs, err := ValueStringRender(chart, `
+spire-server:
+  kubeConfigs:
+    clusterb:
+      externalSecret:
+        name: my-ext-secret
+`)
+			Expect(err).Should(Succeed())
+			Expect(objs[secretTmpl]).ShouldNot(ContainSubstring("kind: Secret"))
+			Expect(objs[serverTmpl]).Should(ContainSubstring("name: my-ext-secret"))
+			Expect(objs[serverTmpl]).Should(ContainSubstring("path: clusterb"))
+		})
+	})
 })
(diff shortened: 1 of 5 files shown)


tests/unit/spire_test.go
```diff
@@ -140,6 +140,39 @@ spire-server:
...
+		// helmutil.CoalesceValues mutates the chart it is given, so each spec loads a
+		// fresh chart to stay independent of the others touching the same value subtree.
+		It("generates the kubeconfigs secret from inline values by default", func() {
+			freshChart, err := helmloader.Load("../../charts/spire")
+			Expect(err).Should(Succeed())
+			objs, err := ValueStringRender(freshChart, `
+spire-server:
+  kubeConfigs:
+    aegis-cluster-dev:
+      kubeConfig: hi
```
kfox1111: Thanks for the contribution.

I'm wondering if we should hang off the secret off of here in the api, rather then make a whole different branch for kubeConfigsExternalSecrets with enables and such.

sabsari: Thanks for taking a look, and good point.

I originally used a single top-level `kubeConfigsExternalSecret` block to mirror
the current shape: there's one `kubeconfigs` volume backed by one Secret, and
consumers subPath-mount `<entryName>` from it. A single external Secret keyed by
entry name drops straight into that structure, so the change stayed minimal
(volume `secretName` + skipping generation) with the consumer mounts untouched.

Hanging the reference off each `kubeConfigs.<name>` entry is cleaner API-wise and
I'm happy to go that way. The part I was weighing is the volume side: once entries
can point at different Secrets (and mix inline + external), a single secret-backed
volume no longer works, and handling that with per-entry volumes / dedup logic
would add more complexity than the feature warrants.

kfox1111: hmm.. Yeah, I'm thinking of it from the api standpoint.

The proposed change, is an all or nothing change. You can use kubeConfig or kubeConfigsExternalSecret

Would there ever be a case where someone would want to use both? or that the user wanted to use multiple secrets? If either of these is true, then it may also be better to weave it into kubeConfig rather then it be its own top level.

sabsari: Makes sense — I'll fold it into the per-entry `kubeConfigs` API instead of a top-level block.
"Multiple secrets" in particular is a real case (per-cluster isolation / independent rotation), and doing it per entry gives inline + external mixing for free. 
Dropping the `enabled` / top-level block.

New shape:

```yaml
kubeConfigs:
  cluster-a:
    kubeConfig: |          # existing
      ...
  cluster-b:
    externalSecret:        # new, alongside kubeConfig / kubeConfigBase64
      name: my-secret      # required
      key: cluster-b       # optional, defaults to the entry name
```

Each entry is exactly one of `kubeConfig` / `kubeConfigBase64` / `externalSecret` (guarded). I'll push an update.

kfox1111 (approved):
LGTM. Thanks! :)

sabsari:
Thanks for the thorough review, @kfox1111 !

kfox1111:
Thanks for working through it and making the chart better. :)

PR #890, opened 2026-07-22
Gateway api support

kfox1111:


faisal-memon:
Follow-up to the Gateway API review: the new `gatewayAPI.backendTLS.enabled` value appears to be documented but not actually used anywhere.

The shared helper only reads `gatewayAPI.backendTLS.caCertificateRefs` and otherwise depends on each chart template to pass a separate boolean for whether to emit `BackendTLSPolicy`, but none of the templates seem to derive that boolean from `gatewayAPI.backendTLS.enabled`. So at the moment this looks like a dead or at least misleading config knob for users.

faisal-memon:
I think the edge-terminated Gateway API path is still broken for several HTTPS backends. In the shared helper, `BackendTLSPolicy` is only emitted when the caller passes `backendTLS=true`, but the new gateway templates for `spiffe-step-ssh`, `spire-server`, and the federation endpoint all hardcode `backendTLS` to `false`.

That means setting `gatewayAPI.tlsSecret` flips those services to `HTTPRoute` edge termination while still forwarding plain HTTP to backends that are serving HTTPS, which should fail at runtime. I think those callers either need to enable backend TLS for the terminated case or derive it from the actual backend protocol.

faisal-memon:
Non-blocking follow-up: the `spire-server` values docs still describe a `tlsSecret`-driven terminated Gateway API mode even though the implementation now hardwires these endpoints to passthrough and the `tlsSecret` key is gone from the values block.

In `charts/spire/charts/spire-server/values.yaml`, both the federation and main `gatewayAPI` comments still say `set tlsSecret => HTTPRoute (edge terminate)`, but the corresponding templates now always render `TLSRoute`. It would be good to update those comments so the values surface matches the actual behavior.

faisal-memon (approved):

PR #894, opened 2026-07-28
Wire unsupportedBuiltInPlugins through to NodeAttestor

pius-grainger:
## Summary
- NodeAttestor gets the same unsupportedBuiltInPlugins passthrough KeyManager already has
- Lets built-in NodeAttestor plugins the chart hasn't added structured values for yet (e.g. azure_msi) be configured today, per the existing comment on unsupportedBuiltInPlugins.nodeAttestor inviting this
- No values.yaml change needed - the placeholder already exists
- No Chart.yaml version bump per CONTRIBUTING.md (maintainers handle releases)

## Test plan
- [x] helm lint charts/spire
- [x] helm template charts/spire (default values) - NodeAttestor: does not render, byte-identical diff against pre-change output
- [x] helm template charts/spire -f <values setting unsupportedBuiltInPlugins.nodeAttestor.azure_msi...> - renders correctly nested under NodeAttestor:
- [x] Deployed to a live EKS cluster running spire-server - azure_msi plugin loads successfully alongside aws_iid/join_token/k8s_psat with no errors, server healthy

### DIFF
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -197,7 +197,7 @@ plugins:
 {{- end }}
 {{- end }}
 
-  {{- if or .Values.nodeAttestor.k8sPSAT.enabled .Values.nodeAttestor.externalK8sPSAT.enabled .Values.nodeAttestor.joinToken.enabled .Values.nodeAttestor.httpChallenge.enabled .Values.nodeAttestor.tpmDirect.enabled .Values.nodeAttestor.awsIID.enabled .Values.nodeAttestor.gcpIIT.enabled .Values.nodeAttestor.x509POP.enabled .Values.spireIdentityExchange.enabled }}
+  {{- if or .Values.nodeAttestor.k8sPSAT.enabled .Values.nodeAttestor.externalK8sPSAT.enabled .Values.nodeAttestor.joinToken.enabled .Values.nodeAttestor.httpChallenge.enabled .Values.nodeAttestor.tpmDirect.enabled .Values.nodeAttestor.awsIID.enabled .Values.nodeAttestor.gcpIIT.enabled .Values.nodeAttestor.x509POP.enabled .Values.spireIdentityExchange.enabled .Values.unsupportedBuiltInPlugins.nodeAttestor }}
   NodeAttestor:
   {{- $clusters := default .Values.kubeConfigs .Values.nodeAttestor.externalK8sPSAT.clusters }}
   {{- if or (eq (.Values.nodeAttestor.k8sPSAT.enabled | toString) "true") (and (eq (.Values.nodeAttestor.externalK8sPSAT.enabled | toString) "true") (gt (len $clusters) 0)) }}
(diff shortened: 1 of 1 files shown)


kfox1111:
This is looking good, but I think it needs to respect the NodeAttestor check. There should be only ever one, even if this feature is used.

pius-grainger:
Good catch, thanks — you were right, and it was worse than cosmetic. I was emitting a second `NodeAttestor:` key *inside* the block that was already open, so when a structured attestor and this escape hatch were both used, the later key won and the structured attestors were **silently dropped**.

Reproduced on the previous revision with `nodeAttestor.joinToken.enabled=true` plus an `unsupportedBuiltInPlugins.nodeAttestor` entry — the rendered `server.conf` contained only:

```json
"NodeAttestor": [
  { "aws_iid": { "plugin_data": { "assume_role": "spire-server-attestation" } } }
]
```

`join_token` and `k8s_psat` were gone.

Fixed by rendering the plugin data into the existing section instead of opening a new one, so there is only ever one `NodeAttestor` block. Same values now give:

```json
"NodeAttestor": [
  { "aws_iid":    { "plugin_data": { "assume_role": "spire-server-attestation" } } },
  { "join_token": { "plugin_data": {} } },
  { "k8s_psat":   { "plugin_data": { "clusters": [ ... ] } } }
]
```

Verified three cases:
- structured + unsupported together — all present in one section (above)
- unsupported only, no structured attestor — section still rendered correctly
- defaults untouched — `helm template` output byte-identical to before the change

Note I deliberately did not add an "exactly one" count/fail here the way the agent chart does in `spire-agent/templates/configmap.yaml`: a SPIRE **server** legitimately runs several node attestors at once (we run `aws_iid`, `gcp_iit`, `join_token` and `k8s_psat` together), so the constraint that applies to the agent doesn't apply here. If you did mean an actual count check on the server side too, say the word and I'll add it.

Also signed off the commit to fix the DCO check.

kfox1111:
Oh. right. server not agent. Will have another look. Thanks! :)

kfox1111:
Isn't this already supported via https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire-lib/templates/_helpers.tpl#L219 ?

pius-grainger:
You're right — and I should have found this before opening the PR. Confirmed by testing unmodified `main` directly:

```
helm template spire . -f vals.yaml
```
with
```yaml
spire-server:
  nodeAttestor:
    joinToken:
      enabled: true
  unsupportedBuiltInPlugins:
    nodeAttestor:
      aws_iid:
        plugin_data: { assume_role: spire-server-attestation }
```

renders, with zero changes from this PR:

```json
"NodeAttestor": [
  { "aws_iid":    { "plugin_data": { "assume_role": "spire-server-attestation" } } },
  { "join_token": { "plugin_data": {} } },
  { "k8s_psat":   { "plugin_data": { "clusters": [...] } } }
]
```

`spire-lib.reformat-and-yaml2json` (`spire-lib.config_merge` → `mustMerge` → `spire-lib.plugins_reformat`) already merges `unsupportedBuiltInPlugins` generically for every plugin type at the dict level before the final list-reformat — so this isn't specific to `keyManager`, it already covers `nodeAttestor` (and everything else) with no code change needed. Also confirmed the same for `azure_msi` specifically, which is the actual thing I needed.

Closing this — thanks for catching it, saved me from merging a genuinely unnecessary patch.

PR #895, opened 2026-07-28
Add ca_bundle_path/ca_bundle_paths support to x509pop NodeAttestor

pius-grainger:
## Summary
- x509pop's structured values only ever supported mode=spiffe (spire-identity-exchange flow) - mode=external_pki has no way to configure ca_bundle_path/ca_bundle_paths, both required by SPIRE's own x509pop plugin for that mode (see https://github.com/spiffe/spire/blob/main/doc/plugin_server_nodeattestor_x509pop.md)
- Adds nodeAttestor.x509POP.caBundlePath (string) and caBundlePaths (array) values, rendered only when set
- spiffe_prefix is now only rendered when mode=spiffe (previously rendered unconditionally, which meant it was emitted even for external_pki configs where it isn't used)
- No Chart.yaml version bump per CONTRIBUTING.md (maintainers handle releases)

## Test plan
- [x] helm template charts/spire (default values) - byte-identical diff against pre-change output
- [x] helm template charts/spire -f <values setting mode=external_pki, caBundlePaths=[...]> - ca_bundle_paths renders correctly, spiffe_prefix correctly omitted
- [x] helm template charts/spire -f <values setting mode=external_pki, caBundlePath=...> - ca_bundle_path (singular) renders correctly

### DIFF
--- a/charts/spire/charts/spire-server/templates/configmap.yaml
+++ b/charts/spire/charts/spire-server/templates/configmap.yaml
@@ -282,7 +282,16 @@ plugins:
     x509pop:
       plugin_data:
         mode: {{ .mode }}
+        {{- if eq .mode "spiffe" }}
         spiffe_prefix: {{ include "spire-server.identity-exchange-spiffe-prefix" $root | quote }}
+        {{- end }}
+        {{- if ne .caBundlePath "" }}
+        ca_bundle_path: {{ .caBundlePath | quote }}
+        {{- end }}
+        {{- if not (empty .caBundlePaths) }}
+        ca_bundle_paths:
+          {{ toYaml .caBundlePaths | nindent 10 }}
+        {{- end }}
         max_intermediates: {{ .maxIntermediates }}
         max_rsa_key_size: {{ .maxRSAKeySize }}
         {{- $cn := "" }}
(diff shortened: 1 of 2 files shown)


kfox1111:
This looks to be a duplicate of https://github.com/spiffe/helm-charts-hardened/pull/856. Could they be merged?

pius-grainger:
You're right, this duplicates #856 — and having compared them, #856 is the better change. Closing this in its favour.

For the record, #856 covers everything here and more:

- it fixes the same underlying bug I hit, where `spiffe_prefix` / `svid_prefix` / `agent_path_template` were rendered unconditionally even in `external_pki` mode — #856 moves them into the `else` branch, which is cleaner than my conditional
- it renders `ca_bundle_path`, same as this PR
- it additionally **mounts** the CA bundle (inline `bundle` or `existingConfigMap`) via `x509pop-configmap.yaml` and `server-resource.yaml`, which this PR did not do at all — without that the configured path doesn't exist in the container, so #856 is the one that actually makes `external_pki` usable end to end
- it adds README docs and unit tests

One small thing #856 doesn't cover, purely optional and not worth blocking on: SPIRE's x509pop plugin also accepts `ca_bundle_paths` (plural) for supplying multiple bundles. Easy follow-up if anyone needs it.

@sylvain-blot-sp — no changes needed from you, just flagging the plural variant in case it's useful later.

PR #899, opened 2026-07-31
Make the external server's downstream RBAC subject configurable

sabsari:
## What

Replace the hardcoded `User: spire-root` RBAC subject used when `externalServer`
is true with a configurable `externalServerSubject` block, so the downstream
Role/ClusterRoleBindings can bind to a User, Group, or ServiceAccount.

## Details

- New `externalServerSubject` on `spire-server`, selected by a `kind`
  discriminator:
  ```yaml
  externalServerSubject:
    kind: User          # User | Group | ServiceAccount
    name: spire-root
    namespace: ""       # ServiceAccount only; empty uses the server namespace
  ```
- `spire-server.subject` renders the matching subject; a helper validates `kind`
  and `fail`s on anything else, mirroring the existing `.Values.kind` guard.
  Only evaluated inside the `externalServer` path, so normal deployments are
  unaffected.
- The ServiceAccount and its token are provisioned out-of-band; the chart only
  creates the RBAC bindings that reference it.
- Regenerated the chart README and added a unit render test.

## Notes

- Backward compatible: defaults (`kind: User`, `name: spire-root`) reproduce the
  previous subject.


charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -412,12 +412,26 @@ The code below determines what connection type should be used.
...
+{{-   end }}
+{{-   $kind }}
+{{- end }}
+
 {{- define "spire-server.subject" }}
 subjects:
 {{-   if .Values.externalServer }}
+{{-     if eq (include "spire-server.external-server-subject-kind" .) "ServiceAccount" }}
+- kind: ServiceAccount
+  name: {{ .Values.externalServerSubject.name }}
```
kfox1111: name and namespace here should be `| quote`ed.

---

This would play very nicely with the user:
"spire://example.com/spire/server", setting the AuthenticationConfiguration on the child cluster to point at the oidc discovery provider from the root

and a kubeconfig that uses the https://github.com/spiffe/k8s-spiffe-workload-jwt-exec-auth plugin.

Maybe we could make the plugin/kubeconfig template available in the chart for use on the root cluster.

sabsari: Done — quoted in the latest push.

---

And your idea is exactly what I need!

What's compelling is that it would fundamentally address the weaknesses shared by the static-credential approaches (both client-cert and static SAT): they are long-lived or non-expiring, and in this architecture they can't be refreshed or hot-reloaded, so any rotation forces a spire-server restart. A short-lived, auto-refreshed SPIFFE JWT-SVID sidesteps all of that.

I also confirmed there's already a published image for the plugin (`ghcr.io/spiffe/k8s-spiffe-workload-jwt-exec-auth`), and the chart already has an init-container + shared emptyDir pattern for injecting plugin binaries, so it should fit cleanly.

I'd be happy to take this on. I'll prototype the plugin injection + kubeconfig template and open a follow-up PR as a starting point for discussion.

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -412,12 +412,26 @@ The code below determines what connection type should be used.
...
+{{-     fail (printf "Unknown externalServerSubject.kind: %s (must be \"User\" or \"ServiceAccount\")" $kind) }}
+{{-   end }}
+{{-   $kind }}
+{{- end }}
+
 {{- define "spire-server.subject" }}
 subjects:
 {{-   if .Values.externalServer }}
+{{-     if eq (include "spire-server.external-server-subject-kind" .) "ServiceAccount" }}
+- kind: ServiceAccount
```
kfox1111, line 428: Oh. one more thing here... the else case has apiGroup but the ServiceAccount part here does not. I think its maybe required?

sabsari: Good to double-check, but apiGroup should stay omitted here.
In RBAC subjects, ServiceAccounts belong to the core API group (apiGroup: ""), while User/Group use `rbac.authorization.k8s.io`.
Setting `apiGroup: "rbac.authorization.k8s.io"` on a ServiceAccount subject is actually rejected by the API server.

kfox1111: I just rereviewed https://kubernetes.io/docs/reference/access-authn-authz/rbac/ and your right. Its not required for ServiceAccounts.

It does mention a 3rd option though we probably should support. kind: Group is also an option that may make sense.

sabsari: Good call — added kind: Group support.

It renders like the User subject (apiGroup: rbac.authorization.k8s.io, no namespace), just with kind: Group, and the kind validation now whitelists all three (User / Group / ServiceAccount). Pushed.

kfox1111 (approved):
LGTM. Thanks! :)

PR #900, opened 2026-07-31
Update spire-identity-exchange for 0.4.0

kfox1111:


faisal-memon:
I think there is one thing we should sort out before merge.

`charts/spire-identity-exchange/Chart.yaml` now sets `appVersion: "v0.5.0"`, but `charts/spire-identity-exchange/README.md` still says `AppVersion: 0.4.0`, still says `Version: 0.1.0`, and the new prose says `As of app version v0.4.0...`.

This is more than just a docs mismatch: `charts/spire-identity-exchange/templates/deployment.yaml` passes `$.Chart.AppVersion` into `spire-lib.image`, so unless `image.tag` is overridden the chart will now deploy the `v0.5.0` image by default. One of these versions is wrong, and as written this risks shipping an unintended runtime image tag.

kfox1111:
Ok. I think I fixed the issues you raised. There is a bigger issue with README's not matching in other charts as well. Opened a separate PR to start addressing that issue.

faisal-memon (approved):

PR #901, opened 2026-07-31
Fix release without --dry-run

kfox1111:


faisal-memon:
I think there is one release-flow regression we should fix before merge.

The new Helm-repo preflight is scoped too broadly. `remote_chart_repositories` walks every `charts/*/Chart.yaml` in the repo, and `require_helm_repositories` fails the release if any remote repo used anywhere in the repo is not configured locally. But later the script only refreshes dependencies for the charts actually involved in the release flow.

That means a release of something like `spire-lib` can now fail just because an unrelated chart elsewhere in the repo depends on `smallstep`, `bitnami`, etc. This makes releases more fragile than necessary compared with the current behavior.

I think the repo check should be scoped to the target chart plus the dependent charts the script is actually going to touch.

PR #905, opened 2026-08-04
Include filterByClassName setting for controller manager

JoelGoh92:
Final part of the solution to close the memory scaling issue reported in https://github.com/spiffe/helm-charts-hardened/issues/851.

**Context**
- Threads the spire-controller-manager `filterByClassName` config field (spiffe/spire-controller-manager#710) through the `spire-server` chart's `controllerManager` and `externalControllerManagers.defaults`, defaulting to `false` so existing deployments are unaffected.
- Combined with the auto class-name label added in #877, operators can opt in to restrict the ClusterSPIFFEID informer cache to a single class, cutting memory usage in nested SPIRE setups where an external controller manager otherwise caches every ClusterSPIFFEID from a child cluster.


charts/spire/charts/spire-server/values.yaml
```diff
@@ -646,6 +638,8 @@ controllerManager:
   className: ""
   ## @param controllerManager.watchClassless specify to process custom resources without class name specified. Useful to slowly migrate to class names from classless installs. Do not have two installs on the same k8s cluster both set to true.
   watchClassless: false
+  ## @param controllerManager.filterByClassName If true, restricts the controller manager's ClusterSPIFFEID informer cache to only objects labeled with this controller's className, using the well-known spire.spiffe.io/class-name label. Requires className to be set. Only enable after confirming all target ClusterSPIFFEIDs already carry that label (this chart labels the ones it renders automatically); enabling it before then will cause the controller manager to see zero matching entries and delete existing registrations. See spire-controller-manager's clusterSPIFFEIDLabelSelector/filterByClassName docs.
```
kfox1111: Looks like the description here is getting too huge and forcing the readme to regenerate the whole doc. Such long doc strings are getting problematic. any way to collapse it a bit to something smaller? Not sure the user needs all the gory details here.

JoelGoh92: Sure, let me update this

JoelGoh92: I've updated the description for this, do let me know if this reads better. Thanks!

kfox1111 (approved):
LGTM. Thanks! :)

PR #906, opened 2026-08-04
Add externalTrafficPolicy support for spire-server LoadBalancer service

dmosyan:
Add a `service.externalTrafficPolicy` value for the spire-server chart, allowing users to set `Local` traffic policy when `service.type` is `LoadBalancer` to preserve client source IP and avoid a second hop. Kubernetes defaults this field to `Cluster` when unset ([docs](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip)). This makes the field configurable instead of hardcoding that behavior.


kfox1111 (approved):
LGTM. Thanks! :)

PR #907, opened 2026-08-05
JWT-SVID exec-auth source for kubeConfigs entries (follow-up to #899)

sabsari:
## What

This is a **draft / for-discussion** follow-up to #899, prototyping the idea @kfox1111 suggested there: authenticate the nested root's access to downstream (child) cluster API servers with short-lived, auto-refreshed SPIFFE **JWT-SVIDs** via a client-go exec credential plugin, instead of static credentials (client-cert / static SAT).

It has two parts, and the second is the reason it's a draft:

- **A chart feature (implemented here):** a new `jwtSVIDExec` source on   `kubeConfigs.<entry>` plus a global `jwtSVIDExecConfig` block. The chart stages the exec plugin binary into the existing shared plugins volume (the #859 init-container pattern), mounts the local agent Workload API socket into the kubeConfigs consumers, and generates an exec-credential kubeconfig.
- **An architecture problem I hit in live testing (needs your call):** wiring exec-auth onto the *external controller-manager sidecar* creates a readiness **deadlock**. Details below — that's the main thing I want to figure out before polishing the rest.

## The values API

```yaml
kubeConfigs:
  clusterd:
    jwtSVIDExec:                                      # 4th exactly-one source
      server: https://clusterd-api.example.com:6443   # target apiserver URL
      certificateAuthorityData: LS0tLS1CRUdJ...        # apiserver CA bundle, base64 PEM
      audience: k8s                                    # optional, default k8s

jwtSVIDExecConfig:                                     # global wiring, shared by all jwtSVIDExec entries
  image:
    registry: ghcr.io
    repository: spiffe/k8s-spiffe-workload-jwt-exec-auth
    tag: "0.1.0"
  pluginPath: /ko-app/cmd
  csiDriverName: csi.spiffe.io
  socketPath: /run/spire/agent-sockets/spire-agent.sock
```

`jwtSVIDExec` is the fourth exactly-one source on a `kubeConfigs` entry (alongside `kubeConfig` / `kubeConfigBase64` / `externalSecret`), presence-activated — no separate enable flag, matching the existing init-container gating. The generated kubeconfig holds no secret; the token is produced at call time by the plugin.

## The deadlock (main discussion point)

When I designed this, I was already worried there might be a cycle here. So instead of just reasoning about it, I built it to see what actually happens — and it doesn't work as-is.

Two kinds of container consume `/kubeconfigs/<entry>`:

- the **spire-server container** itself (`k8s_psat` / `k8s_bundle` notifier / `bundlePublisher`) — **safe**: its readiness doesn't depend on issued identity.
- the **external `spire-controller-manager-<child>` sidecar** — **this is the problem.**

The sidecar runs inside the server pod. To fetch a JWT-SVID the exec plugin needs the accessor identity, which the local agent only has after it has synced from the server. At cold boot (or any transient identity race) the fetch fails, the sidecar can't reach the child cluster and CrashLoops, which makes the **whole pod NotReady**. Because the
spire-server `Service` is readiness-gated, a NotReady pod is dropped from its endpoints — so **every agent loses the server**, identities stop syncing, and the sidecar never gets its identity.

## Where I'd like your input

There are ways to work around it — for example `publishNotReadyAddresses: true` on the server `Service`, or raising the controller-manager's `cacheSyncTimeout` so a cold-boot race doesn't hit the fatal exit — but neither is a real fix. They just hide the cycle instead of breaking it.

It's also very possible the architecture I prototyped here is just the wrong approach. If you have ideas, I'd be happy to talk them through together.

## Notes

- Smoke-tested live end-to-end (root pod → exec plugin → fresh JWT-SVID → child apiserver authn → RBAC) with zero static tokens: the **server-container** consumer path works; the **sidecar** path is where the deadlock surfaced.
- Backward compatible: with no `jwtSVIDExec` entries, none of the new wiring renders.


kfox1111:
This is only a problem when not using the (root -> external server) -> child cluster pattern I think, which is the recommended way....

But, there is probably a simpler way of doing this....

The controller-manager already needs access to the spire-server admin socket. It wouldnt be too hard to extend https://github.com/spiffe/k8s-spiffe-workload-jwt-exec-auth to support getting the jwt from there as well as from the workload api. Then the controller-manager wouldn't depend on csi/workload socket/spire-agent of the underlying server, and wouldn't need any new permissions.

sabsari:
You're right that this only happens in that specific architecture. But it's still worth solving.

Your spire-server admin socket idea makes the problem go away in every case. I'll start by extending k8s-spiffe-workload-jwt-exec-auth, then rework this on top of it.

sabsari:
Reworked on top of the `server-admin-api` source (now merged and released as
`ghcr.io/spiffe/k8s-spiffe-workload-jwt-exec-auth:0.2.0`). The plugin mints the accessor
JWT-SVID from the SPIRE server admin socket, so the cold-boot mint deadlock is gone.

Values API now:

```yaml
kubeConfigs:
  clusterd:
    jwtSVIDExec:                                      # 4th exactly-one source
      server: https://clusterd-api.example.com:6443   # target apiserver URL
      certificateAuthorityData: XXXXXXXX              # apiserver CA bundle, base64 PEM
      audience: k8s                                   # optional, default k8s

jwtSVIDExecConfig:                                    # global wiring
  image:
    registry: ghcr.io
    repository: spiffe/k8s-spiffe-workload-jwt-exec-auth
    tag: "0.2.0"
  pluginPath: /ko-app/cmd
  spiffeID: spiffe://example.org/spire-root           # accessor identity the plugin mints
```

One requirement worth flagging: the root server must use a **persistent `keyManager`**
(`disk`, AWS KMS, …), not in-memory. The child validates these JWT-SVIDs against the root's
OIDC JWKS, so the JWT signing key has to survive restarts — an in-memory key rotates on every
restart and breaks validation until the new key propagates.

kfox1111:
> One requirement worth flagging: the root server must use a persistent keyManager
(disk, AWS KMS, …), not in-memory. The child validates these JWT-SVIDs against the root's
OIDC JWKS, so the JWT signing key has to survive restarts — an in-memory key rotates on every
restart and breaks validation until the new key propagates.

Alternately, it can use the spire-server-attestor-tpm (https://github.com/spiffe/spire-server-attestor-tpm) plumbing to fetch up to date trust bundles. I've been meaning to add chart support for it at some point, but its not currently how I'm using the chart (I use the examples/bottom-turtle-ha like architecture) so its been lower on the todo list right now.

charts/spire/charts/spire-server/templates/_helpers.tpl
```diff
@@ -186,6 +186,60 @@ Name of the chart-generated Secret holding the inline kubeConfigs entries.
...
+
+{{- define "spire-server.jwt-svid-exec-kubeconfig" -}}
+{{- $jwtSVIDExec := .jwtSVIDExec -}}
+{{- $root := .root -}}
+{{- $spiffeID := $root.Values.jwtSVIDExecConfig.spiffeID -}}
+{{- if not $spiffeID -}}
+{{- fail "jwtSVIDExecConfig.spiffeID is required when a kubeConfigs entry uses jwtSVIDExec" -}}
+{{- end -}}
+{{- if not (hasPrefix "spiffe://" $spiffeID) -}}
+{{- fail (printf "jwtSVIDExecConfig.spiffeID %q must be a spiffe:// URI" $spiffeID) -}}
```
kfox1111, line 205: how would you feel about having it prepend spiffe://<trustdomain> to a spiffeID that starts with "/"? then we could default the value to something like "/spire-root" and it should "just work" for most folks out of the box? One less complicated thing for a user to figure out.

sabsari: Done. spiffeID now takes a "/" prefixed path that's expanded with the chart trust domain, defaulting to "/spire-root". Thanks, much nicer out of the box.

kfox1111 (approved):
LGTM. Thanks! :)

sabsari:
Thanks for merging!

On the persistent-keyManager point: besides the `spire-server-attestor-tpm` plumbing you suggested, I think it might also be solved by feeding the OIDC provider from the bundle ConfigMap (`spiffe-oidc-discovery-provider.bundleSource: ConfigMap`) instead of the agent, so JWKS stays fresh independently of agent sync. I'll test both and follow up.

PR #909, opened 2026-08-05
Add PodDisruptionBudget support to spire-server

### LINKED ISSUE #577: Add configurable PodDisruptionBudget for spire-server
It would be great to be able to configure a PodDisruptionBudget in the helm chart. This is relevant for spire-server in HA mode with > 1 replicas.

dmosyan:
Add PodDisruptionBudget configuration support to protect `spire-server` pods during cluster disruptions. 
Fixes #577


charts/spire/charts/spire-server/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: {{ include "spire-server.fullname" . }}
+  namespace: {{ include "spire-server.namespace" . }}
+  labels:
+    {{- include "spire-server.labels" . | nindent 4 }}
+spec:
+  {{- if .Values.podDisruptionBudget.maxUnavailable }}
+  maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
+  {{- else }}
+  minAvailable: {{ .Values.podDisruptionBudget.minAvailable | default 1 }}
```
kfox1111: should we default this to `(sub .Values.replicaCount 1)` ? Then it automatically scales with the users setting.

We should figure out how to do the same kind of poddisruptionbudget in charts/spire-nested/templates/ for tag haAgentCommon too. To make sure one side of the ha pair is always available...

dmosyan: @kfox1111 thanks for the review. 

> should we default this to `(sub .Values.replicaCount 1)` ? Then it automatically scales with the users setting.

That's a good point. But if a user sets `replicaCount: 1` the `maxUnavailable` will be 0, which blocks the cluster maintenance. I think it's fine. 

Another option is to create PDB if enabled and `replicaCount` > 1 + your suggestion ` maxUnavailable: {{ sub .Values.replicaCount 1 }}`. Let me know which one your think is better approach. 

> We should figure out how to do the same kind of poddisruptionbudget in charts/spire-nested/templates/ for tag haAgentCommon too. To make sure one side of the ha pair is always available...

I implemented the changes for `spire-ha-agent`. As it's a DaemonSet I added `maxUnavailable: 1` (one node at a time). Let me know if you think another approach is better. If this needs further discussion I can remove it from this PR and open an issue to discuss the solution first.

marcofranssen: On very large clusters a PDB of 1 for spire-agent would mean it will take ages for a helm install to finish. We deliberately do many in same time, like 25% to ensure our upgrades are done more quickly

dmosyan: Makes sense. Is it aligned with the minimum availability requirement for `spire-ha-agent`? Do you want a separate issue/PR?

charts/spire-nested/templates/poddisruptionbudget-ha-agent.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if and .Values.tags.haAgentCommon .Values."spire-ha-agent".podDisruptionBudget.enabled }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: spire-ha-agent
```
kfox1111: Sorry, I think I didnt word very well my intention here.

In the ha agent deployment mode, there are two spire-server instances, a and b. Its the spire-servers that should have the pdb on it. so only server a or server b could be offline, not both.

charts/spire/charts/spire-server/templates/poddisruptionbudget.yaml
kfox1111: yeah, replicaCount - 1 but minimum of 1 makes sense to me. If the user wants to force min 0 they still can.

kfox1111: for the spire-ha-agent stuff, see comment below.

dmosyan: Thanks for clarifying. I misunderstood it initially. I have reverted the `spire-ha-agent` changes and set the correct default PDB value for `spire-server`. Is anything else you want to adjust?

kfox1111 (approved):
LGTM. Thanks! :)

dmosyan:
@kfox1111 thanks for the review and for the feedback.

fastly-ffej (approved):

PR #913, opened 2026-08-12
fix(spire-agent): suffix spire-config ConfigMap name per agent profile

### LINKED ISSUE #912: spire-agent: spire-config ConfigMap volume not suffixed for additional agent profiles
## Description

The `spire-agent` chart's multi-DaemonSet support (`spire-agent.agents.<name>`) correctly suffixes every per-profile resource — the profile's own ConfigMap, the trust-bundle ConfigMap volume, and the DaemonSet name itself — with `$nameSuffix` (e.g. `-<name>`). The one exception is the DaemonSet's `spire-config` volume, which mounts `agent.conf` and is hardcoded to the **default** profile's ConfigMap regardless of which profile the DaemonSet belongs to:

https://github.com/spiffe/helm-charts-hardened/blob/main/charts/spire/charts/spire-agent/templates/daemonset.yaml#L474-L476

```yaml
volumes:
  - name: spire-config
    configMap:
      name: {{ include "spire-agent.fullname" . }}   # missing $nameSuffix
```

## Impact

Any additional `agents.<name>` profile silently mounts the **default** profile's `agent.conf` instead of its own — whatever that profile's actual config sets (`nodeAttestor`, `workloadAttestors`, `customPlugins`, socket paths, anything) is never read, because the profile's own correctly-rendered ConfigMap is never mounted anywhere. This applies to any use of the multi-agent feature where a profile's config differs from the default, not to any one workload-attestor setting in particular.

We hit this running a second profile with a different `workloadAttestors.k8s` configuration than the default — the setting appeared correctly in the rendered ConfigMap but never took effect at runtime. Confirmed via live pod spec inspection that the profile's DaemonSet was mounting `spire-agent` (the default ConfigMap), not `spire-agent-<profile>`.

## Reproduction

1. Configure `spire-agent.agents.<name>` with any profile-specific config that differs from the default.
2. `helm template` (or deploy) and inspect the profile's DaemonSet: `spec.template.spec.volumes[] | select(.name == "spire-config")`.
3. Its `configMap.name` reads the default profile's fullname, not `<fullname>-<name>`.

## Fix

One-line fix — add the missing `$nameSuffix`, matching the pattern already used for every other per-profile resource in this same template:

```diff
-            name: {{ include "spire-agent.fullname" . }}
+            name: {{ printf "%s%s" (include "spire-agent.fullname" .) $nameSuffix | quote }}
```

Confirmed present on current `main` as well as older `0.26.1`. PR open: #913.

scubadam:
Fixes #912

## What
The `spire-agent` DaemonSet's `spire-config` volume (which mounts `agent.conf`) is hardcoded to the default profile's ConfigMap name, unlike every other per-profile resource in this template (the profile's own ConfigMap, the trust-bundle volume, the DaemonSet name itself), which all correctly append `$nameSuffix`.

```diff
-            name: {{ include "spire-agent.fullname" . }}
+            name: {{ printf "%s%s" (include "spire-agent.fullname" .) $nameSuffix | quote }}
```

## Why
Any `spire-agent.agents.<name>` additional profile silently mounts the default profile's `agent.conf` instead of its own — whatever that profile's config actually sets (`nodeAttestor`, `workloadAttestors`, `customPlugins`, socket paths, anything) is never read, since the profile's own correctly-rendered ConfigMap is never mounted anywhere. This is independent of what a given profile's config contains — it breaks the multi-agent feature generally, for any profile whose config differs from the default.

Found this running a second profile with different `workloadAttestors.k8s` config than the default. The config rendered correctly in the ConfigMap but never took effect at runtime. Confirmed via live pod spec inspection that the profile's DaemonSet was mounting `spire-agent` (default), not `spire-agent-<profile>`.

## Testing
- `helm template` confirms the additional profile's DaemonSet now references its own correctly-suffixed ConfigMap, matching the DaemonSet name and the trust-bundle volume's existing (already-correct) behavior.
- No `Chart.yaml`/`values.yaml` changes, so no README regen needed.
- Verified against a real cluster running this chart with a second agent profile — the profile-specific config is now actually read after this fix.

### DIFF
--- a/charts/spire/charts/spire-agent/templates/daemonset.yaml
+++ b/charts/spire/charts/spire-agent/templates/daemonset.yaml
@@ -473,7 +473,7 @@ spec:
       volumes:
         - name: spire-config
           configMap:
-            name: {{ include "spire-agent.fullname" . }}
+            name: {{ printf "%s%s" (include "spire-agent.fullname" .) $nameSuffix | quote }}
         {{- if .Values.keyManager.disk.enabled }}
         - name: spire-key-manager
           {{- if eq .Values.keyManager.disk.mode "hostPath" }}


kfox1111 (approved):
LGTM. Thanks! :)

PR #918, opened 2026-08-17
fix(spire): preserve webhook order in patch hooks

RuriRyan:
## Summary
All install/upgrade hook patches preserve the canonical federation-then-SPIFFE-ID webhook order. When install/upgrade hooks are disabled, the rendered ValidatingWebhookConfiguration uses the configured failure policy and no lifecycle Jobs are rendered.

## Testing
- Targeted Ginkgo render tests for webhook order and hook-disabled behavior
- Full Ginkgo unit suite
- `helm lint charts/spire`


kfox1111:
I would have thought GitOps shops would disable those hooks, as hooks are not so GitOps friendly?

I dont think document ordering in helm is guaranteed either? Could just the ordering of docs getting rendered also cause problems?

RuriRyan:
Arrays are always ordered in Go, maps aren't. As long as those don't get mixed up in some parsing the order is preserved.
Also found another issue, where the failurePolicy wasn't updated even if the install and upgrade hooks were disabled globally. It stayed on Ignore.

kfox1111 (approved):
Not sure if this will always work in all cases, but I don't see any harm in it either. If it works reliably for you, that is valuable all on its own.

LGTM. Thanks! :)

tests/unit/spire_test.go
```diff
@@ -663,6 +730,84 @@ spire-server:
...
+				"spire/charts/spire-server/templates/post-install-hook.yaml",
+				"spire/charts/spire-server/templates/pre-upgrade-hook.yaml",
+				"spire/charts/spire-server/templates/post-upgrade-hook.yaml",
+			} {
+				documents, err := decodeRenderedDocuments(objs[template])
+				Expect(err).Should(Succeed())
+				for _, document := range documents {
+					Expect(document.Kind).ShouldNot(Equal("Job"), template)
+				}
+			}
```
kfox1111, line 810:
```suggestion
			}
		})
	})
```

kfox1111: github's rebase tool munged the merge. missed some closing brackets

PR #920, opened 2026-08-17
fix(spiffe-oidc-discovery-provider): run under restricted PSA/SCC on OpenShift

### LINKED ISSUE #919: OIDC provider needlessly forces spire-server to privileged PSA + SCC on OpenShift
## Summary

On OpenShift, enabling the `spiffe-oidc-discovery-provider` (the default) causes the chart to:

1. **Downgrade the `spire-server` namespace from `restricted` to `privileged` Pod Security Admission**, and
2. **Create a fully privileged `SecurityContextConstraints`** bound to the OIDC provider's ServiceAccount.

Both stem from a single incorrect assumption — that the OIDC discovery provider needs elevated privileges. It does not. Live evidence shows the pod runs happily under OpenShift's built-in `restricted-v2` SCC, fully hardened, using only `restricted`-compatible volumes.

This contradicts the chart's own [Namespaces documentation](https://spiffe.io/docs/latest/spire-helm-charts-hardened-about/namespaces/), which instructs labeling `spire-server` with `pod-security.kubernetes.io/enforce=restricted`, and it forces a workaround (creating `spire-server` outside the chart) to keep the namespace hardened.

## Environment

- Chart: `spire` / `spire-nested` `0.30.0`, subchart `spiffe-oidc-discovery-provider` `0.1.0`, appVersion `1.15.2`
- Platform: OpenShift / ROSA (HCP) 4.x
- `global.openshift: true`, `spiffe-oidc-discovery-provider.enabled: true` (default)

## Defect A — namespace PSA downgrade

`charts/spire-lib/templates/_spire-server-namespace.yaml`:

```gotmpl
{{- if (dig "openshift" false .Values.global) }}
{{-   $_ := set $labels "security.openshift.io/scc.podSecurityLabelSync" "false" }}
{{-   if (index .Values "spiffe-oidc-discovery-provider").enabled }}
{{-     $_ := set $labels "pod-security.kubernetes.io/enforce" "privileged" }}
{{-   end }}
{{- end }}
```

When the chart creates `spire-server` on OpenShift and the OIDC provider is enabled (the default), it sets `pod-security.kubernetes.io/enforce: privileged` on the namespace. This directly contradicts the [Namespaces docs](https://spiffe.io/docs/latest/spire-helm-charts-hardened-about/namespaces/):

> ```shell
> kubectl create namespace spire-server
> kubectl label namespace spire-server pod-security.kubernetes.io/enforce=restricted
> ```

So the documentation says `spire-server` should be `restricted`, while the template makes it `privileged`.

## Defect B — privileged SCC

`charts/spiffe-oidc-discovery-provider/templates/scc-spire-oidc-discovery-provider.yaml` (gated only by `global.openshift`, no opt-out value):

```yaml
runAsUser:
  type: RunAsAny
seLinuxContext:
  type: RunAsAny
supplementalGroups:
  type: RunAsAny
volumes:
  - configMap
  - csi
  - downwardAPI
  - emptyDir
  - ephemeral
  - hostPath
  - projected
  - secret
allowHostDirVolumePlugin: true
allowHostIPC: true
allowHostNetwork: true
allowHostPID: true
allowHostPorts: true
allowPrivilegeEscalation: true
allowPrivilegedContainer: true
fsGroup:
  type: RunAsAny
seccompProfiles:
  - '*'
```

This grants host namespace access, privileged containers, `hostPath`, arbitrary seccomp, and `RunAsAny` to the ServiceAccount of an internet-facing OIDC endpoint.

## Evidence both are unnecessary

The pod runs under `restricted-v2`, fully hardened, and uses none of the above:

```console
$ oc -n spire-server get pod <oidc-pod> \
    -o jsonpath='{.metadata.annotations.openshift\.io/scc}'
restricted-v2

# container securityContext (OpenShift-assigned):
runAsNonRoot: true
runAsUser: 1001310000
allowPrivilegeEscalation: false
readOnlyRootFilesystem: true
capabilities: { drop: [ALL] }
seccompProfile: { type: RuntimeDefault }
```

Supporting facts:

- The chart's own default security context (`spire-lib.securitycontext`) is already hardened identically (`allowPrivilegeEscalation: false`, `runAsNonRoot: true`, `readOnlyRootFilesystem: true`, `drop: [ALL]`, `RuntimeDefault`).
- The deployment mounts only `csi` (the `csi.spiffe.io` inline workload-API volume), `configMap`, `emptyDir`, `secret`, `projected`, and `downwardAPI` — all permitted under `restricted-v2`. The SCC's `hostPath` grant is never used.
- The container image is `nginxinc/nginx-unprivileged` — designed to run without privilege.

## The workaround this forces

To keep `spire-server` hardened, we cannot let the chart create it. We create it manually with `restricted` PSA and use the two-namespace configuration (`namespaces.create: false`, `namespaces.system.create: true`). This is a **workaround**, not a preferred configuration:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: spire-server
  labels:
    pod-security.kubernetes.io/enforce: restricted
    pod-security.kubernetes.io/audit: restricted
    pod-security.kubernetes.io/warn: restricted
    # (scc.podSecurityLabelSync=false handled as the chart does)
```

With this, the OIDC provider correctly lands on `restricted-v2` and the namespace stays `restricted` — confirming the chart's privileged PSA/SCC are not needed.

## Security impact

- An internet-facing OIDC discovery endpoint's namespace and ServiceAccount are needlessly granted host access and privileged execution.
- Violates least privilege and undermines the premise of the "hardened" chart.
- Contradicts the chart's own Namespaces documentation.
- Proliferates an unnecessary custom SCC that trips cluster security policies and audits.

## Expected behavior

- The OIDC discovery provider should run under `restricted-v2` (or the platform default).
- The `spire-server` namespace should remain `pod-security.kubernetes.io/enforce=restricted`, per the docs — no `privileged` override.
- No custom privileged SCC should be created for the OIDC provider. If any SCC is retained, it should be minimal (nonroot, no host access, no privilege escalation, no `hostPath`, `MustRunAsRange` instead of `RunAsAny`) and opt-in.

## Suggested fixes

1. Remove the `pod-security.kubernetes.io/enforce: privileged` branch for the OIDC provider in `_spire-server-namespace.yaml`.
2. Drop the OIDC provider SCC on OpenShift (rely on `restricted-v2`), or gate it behind an explicit value defaulting to off, or strip all host/privileged grants and `hostPath`.
3. Keep the namespace PSA and the OIDC provider's privileges consistent with the Namespaces documentation.

## Reproduction

1. Install on OpenShift with defaults and let the chart create `spire-server`:
   `global.openshift: true`, `global.spire.namespaces.server.create: true`, `spiffe-oidc-discovery-provider.enabled: true`.
2. Observe the namespace is forced to privileged:
   ```console
   $ oc get ns spire-server -o jsonpath='{.metadata.labels.pod-security\.kubernetes\.io/enforce}'
   privileged
   ```
3. Observe the pod actually runs under `restricted-v2`, proving the privilege is unused:
   ```console
   $ oc -n spire-server get pod <oidc-pod> -o jsonpath='{.metadata.annotations.openshift\.io/scc}'
   restricted-v2
   ```

Munken:
## What

On OpenShift, running the `spiffe-oidc-discovery-provider` no longer requires elevated privileges:

- Removes the `pod-security.kubernetes.io/enforce: privileged` override that was applied to the `spire-server` namespace whenever the OIDC provider was enabled. The namespace now stays `restricted` (the `security.openshift.io/scc.podSecurityLabelSync=false` label is retained).
- Deletes the fully privileged `SecurityContextConstraints` created for the OIDC provider's ServiceAccount. The provider falls through to OpenShift's built-in `restricted-v2`.

## Why

The OIDC discovery provider runs fine under `restricted-v2`: non-root, no privilege escalation, all capabilities dropped, `RuntimeDefault` seccomp, read-only root filesystem, and it mounts only restricted-compatible volumes (`csi`, `configMap`, `emptyDir`, `secret`, `projected`, `downwardAPI`). The previous privileged SCC (host network/IPC/PID, privileged container, `hostPath`, arbitrary seccomp, `RunAsAny`) and the privileged namespace PSA were unused, contradicted the chart's own [Namespaces documentation](https://spiffe.io/docs/latest/spire-helm-charts-hardened-about/namespaces/) (which specifies `restricted` PSA for `spire-server`), and violated least privilege for an internet-facing endpoint.

Fixes #919

## How tested

`helm template` of the `spire` chart with `global.openshift=true` and the OIDC provider enabled:

- `spire-server` namespace renders `pod-security.kubernetes.io/enforce=restricted` (previously `privileged`), with `scc.podSecurityLabelSync=false` retained.
- No `spiffe-oidc-discovery-provider` SCC is emitted.
- The OIDC provider Deployment still renders with a hardened container securityContext (`allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `readOnlyRootFilesystem: true`, `runAsNonRoot: true`, `RuntimeDefault`) — compatible with `restricted-v2`.
- With the OIDC provider disabled, the namespace remains `restricted` (unchanged).

`helm lint` passes on the `spire` and `spire-lib` charts.


kfox1111:
@mrsabath This pr seems good to me. Could you please verify?

Munken:
Rebased with sign-off

Munken:
> @mrsabath This pr seems good to me. Could you please verify?

I experienced some problems with this today. I’ll run some more investigations and report back.

Munken:
Alright problem understood. 

When installing the chart (with modified namespace) using Argo I was seeing:
```
one or more synchronization tasks completed unsuccessfully, reason: statefulsets.apps "spire-internal-server" is forbidden: ose-spire-internal-server uses an inline volume provided by CSIDriver upstream.csi.spiffe.io and namespace ose-spire-server has a pod security enforce level that is lower than privileged
```

However, this turned out to be a race condition. The reason is that openshift checks the `security.openshift.io/csi-ephemeral-volume-profile` label at admission time. But as the CSI driver when not fully installed at admission time, then this label was not available and treated as `priviledged`.

My suggestion would be to add a sync wave annotation on the CSI driver. I've added it to this PR.

kfox1111:
> My suggestion would be to add a sync wave annotation on the CSI driver. I've added it to this PR.

Hmmm... Feels like this is unrelated to OpenShift. Like, this would hit anyone using argocd and maybe other cd software like flux.

Should we make the whole annotations configurable? Is -1 always the right value or should it be possible for the user to override it to a bigger negative yet (maybe when spire chart is nested in another chart?)

Munken:
> > My suggestion would be to add a sync wave annotation on the CSI driver. I've added it to this PR.
> 
> Hmmm... Feels like this is unrelated to OpenShift. Like, this would hit anyone using argocd and maybe other cd software like flux.
> 
> Should we make the whole annotations configurable? Is -1 always the right value or should it be possible for the user to override it to a bigger negative yet (maybe when spire chart is nested in another chart?)

It is hitting openshift because it uses security.openshift.io/csi-ephemeral-volume-profile to determine which CSIDrivers are safe. As far as I know there is nothing equivalent in the other managed kubernetes. So the requirement for the csidriver to come first is openshift specific. 

I can make the annotation and value configurable.

Munken:
Added general support for annotating the csiDriver and explicitly handle the sync wave.

Munken:
I also looked into flux. It seems to me that here is no simple equivalent.
https://fluxcd.io/flux/components/kustomize/kustomizations/#kustomization-dependencies

kfox1111 (approved):
LGTM. Thanks! :)

PR #923, opened 2026-08-18
Allow sqlite3 in memory when kind is deployment

daescha:
we use in memory sqlite3 database and have no use for the statefulset, all it does is occasionally cause problems for us on an upgrade. Would like to use the deployment option. this mr adds values key for indicating in memory is in use and allows deployment if sqlite3 is set and inMemory is true.


kfox1111:
Curious how this is working....

Are you:
* manually orchestrating so that spire-agents are stateless and recreated after the deployment pods update?
* Relying on agent reattestation  to reestablish trust automatically?
* using federation?

daescha:
@kfox1111 we rely on agent reattestation, no federation, no manual orchestration. heres a condensed version of the values.yaml we typically use as a base:

```yaml
spire-server:
  persistence:
    type: emptyDir
  dataStore:
    sql:
      file: memdb
      options:
        - mode: memory
        - cache: shared
  keyManager:
    disk:
      enabled: false
    memory:
      enabled: false  # true only when no KMS/vault upstream auth configured (i.e., a test deployment)
  nodeAttestor:
    k8sPSAT:
      serviceAccountAllowList:
        - "spire-system:spire-agent"
  controllerManager:
    enabled: true
    staticManifestMode: internal
    reconcile:
      clusterSPIFFEIDs: false
      clusterStaticEntries: true
      clusterFederatedTrustDomains: false
    identities:
      clusterStaticEntries:
        agent:
          parentID: "spiffe://example.org/spire/server"
          spiffeID: "spiffe://example.org/ns/spire-system/sa/spire-agent"
          selectors:
            - "k8s_psat:cluster:my-cluster"
            - "k8s_psat:agent_ns:spire-system"
            - "k8s_psat:agent_sa:spire-agent"
        my-workload:
          parentID: "spiffe://example.org/ns/spire-system/sa/spire-agent"
          spiffeID: "spiffe://example.org/ns/my-ns/sa/my-workload"
          selectors:
            - "k8s:ns:my-ns"
            - "k8s:sa:my-workload"
            - "k8s:pod-label:app.kubernetes.io/name:my-workload"
        # ~10 more of these
spire-agent:
  rebootstrapDelay: 15s
```

kfox1111:
Hmm... so when combined with an upstream ca, you dont see a split brain situation.... only the inability to issue jwts until reattestation kicks in... That seems safe enough...

There are probably other cases that are unsafe in this config... Like the testing one you mention, but only for testing, that is fine... We should see if there are any warnings or failures we could give out in the edge configs that could bite people though.

daescha:
@kfox1111 Added warnings in NOTES.txt for the combinations that bite, all gated on
`inMemory` so nobody else sees them:

* no controller-manager reconciler enabled: entries live only in memory and
  nothing recreates them, so they're lost on restart
* `keyManager.memory` with no upstreamAuthority: new CA every restart, the
  whole trust domain re-attests and old SVIDs stop verifying

Kept them as warnings rather than failures since they are reasonable for
testing. Im fine with turning the first into a hard failure if you prefer, doesnt affect me.

added a hard failure when using an in-memory datastore on a deployment that can
surge, since the default rolling update runs two servers with separate datastores
and there's no legitimate reason to want that. It accepts either
`{type: Recreate}` or `{rollingUpdate: {maxSurge: 0}}`; thanks for merging #924,
this needed that value to exist.

also, fwiw, we use `jwtSVIDSupport: false`, so arent affected by the jwt part of your message.

kfox1111 (approved):
LGTM. Thanks! :)

PR #924, opened 2026-08-18
Make spire-server rollout strategy configurable

daescha:
this is related to #923 - useful for if running in memory database with a deployment mode spire server.


charts/spire/charts/spire-server/values.yaml
```diff
@@ -23,6 +23,9 @@ image:
 ## @param kind Define SPIRE server deployment type. Can be statefulset/deployment. Defaults to statefulset if not set. This feature is experimental.
 kind: statefulset
 
+## @param strategy [object] Rollout strategy for the server, mapped to spec.strategy when kind is "deployment" and to spec.updateStrategy when kind is "statefulset". Left empty the Kubernetes default applies, which for a Deployment surges a second server before the old one goes away. Set `{type: Recreate}` to keep at most one server running, as a memory keyManager or an in-memory datastore requires.
+strategy: {}
```
kfox1111: This pr generally looks good to me. Thanks for submitting it.

I think the statefulset name of updateStrategy is a little clearer in purpose then just strategy. Could we rename it to updateStrategy for values?

kfox1111 (approved):
LGTM. Thanks! :)

PR #925, opened 2026-08-19
:sparkles: Add topologySpreadConstraints support to OIDC discovery provider

Munken:
## What changed
Added optional `topologySpreadConstraints` support to the `spiffe-oidc-discovery-provider` Deployment.

- New `topologySpreadConstraints: []` value (default empty, no behavioral change).
- Rendered into the Deployment pod spec via the standard `{{- with .Values.topologySpreadConstraints }}` block.
- Regenerated the chart README param table.

## Why
The OIDC discovery provider previously supported `nodeSelector`, `affinity`, and `tolerations` but not `topologySpreadConstraints`. This brings it in line with `spire-server` and the `spike-*` charts, enabling pod spread for resilience across zones/nodes.

## How tested
- `helm template` with a sample constraint renders it correctly in the pod spec.
- Ran `go test ./...` in `tests/unit`; the only failures are pre-existing on `main` (spire-server updateStrategy/dataStore) and unrelated to this change.
- Regenerated docs via readme-generator.


kfox1111 (approved):
LGTM. Thanks! :)

PR #929, opened 2026-08-22
Bottom turtle ha registry tests

kfox1111:


examples/bottom-turtle-ha/run-tests.sh
```diff
@@ -441,3 +685,12 @@ if [ "${BROKER}" -eq 1 ]; then
   run_federation_test_job
 fi
 
+# The image pull path has to survive losing a side too. Everything it depends on is HA:
+# the node's identity comes from the host spire-ha-agent, and the credential provider
+# talks to the combined exchange endpoint rather than either side directly. Delete the
+# job first so this is a genuine second pull rather than a cached result.
+kubectl delete job image-pull
```
faisal-memon, line 692: The second pull happens after deleting only the in-cluster `spire-a` release. The host `spire-agent@a` and `spire-ha-agent@main`, which provide the kubelet credential provider identity, are still running, so this does not actually exercise HA-agent/root-agent failover. It proves the B-side exchange endpoint remains available, but not the broader HA claim in the comment. Could we stop or isolate the A-side host agent before this pull if that is the behavior we want covered?

faisal-memon:
One CI follow-up: `.github/workflows/helm-chart-ci-ignore.yaml` does not exclude `examples/**/*.sh` or `.github/scripts/install-image-cred-provider.sh`. Changes to either run the “Skipping …” workflow alongside real CI, producing duplicate check names. Could we mirror the primary workflow’s relevant path list there so the bypass workflow only runs when the main workflow is genuinely skipped?

faisal-memon (approved):

faisal-memon (approved):

PR #930, opened 2026-08-24
fix(spire-server): scrape controller-manager metrics port (pm-cm)

Munken:
## Problem
The spire-server `PodMonitor` hardcodes `- port: prom-cm`, but the controller-manager container exposes its Prometheus port as **`pm-cm`** (see `_controller-manager-container.tpl`, `printf "pm-cm%s" .portSuffix`). For external controller managers the port is suffixed per cluster (`pm-cm-<cluster>`, with numeric-preserving/hashed truncation for long names).

Because no pod exposes a `prom-cm` port, the controller-manager metrics (controller-runtime reconcile, workqueue, leader election, certwatcher, rest_client) are **never scraped** when `telemetry.prometheus.podMonitor.enabled=true`. Only the server's own `prom` (9988) endpoint is collected.

## Fix
- Add a `spire-controller-manager.prometheus-port-names` helper that returns the CM Prometheus port names, mirroring the exact iteration and `portSuffix` logic used by `spire-controller-manager.containers` (primary CM plus one entry per `externalControllerManagers`/`kubeConfigs` cluster). It is gated on the same telemetry/recommendations condition and honours `prometheusPortName` overrides.
- Rewrite the PodMonitor's `podMetricsEndpoints` to always scrape `prom` and then `range` over the helper, so scrape endpoints stay in sync with the actual container ports and no longer reference the non-existent `prom-cm`.

## Tests
Added unit tests in `tests/unit/spire_test.go`:
- single CM → PodMonitor scrapes `pm-cm`, not `prom-cm`;
- external CM (`kubeConfigs.microserv`) → PodMonitor scrapes both `pm-cm` and the suffixed `pm-cm-microserv` (matching the rendered container port);
- CM disabled → only the server `prom` port is scraped.

Verified with `helm template` (single, multi-cluster, and long-name hash-suffix cases) and `go test ./tests/unit` (51 passed).

## Impact
No values API change; behaviour only differs when the PodMonitor is enabled. Existing installs gain the previously-missing controller-manager metrics.

### DIFF
--- a/charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
+++ b/charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
@@ -189,3 +189,60 @@ Auto-generation preserves trailing numbers from cluster names or uses hash for u
     {{- toYaml .Values.extraVolumeMounts | nindent 4 }}
     {{- end }}
 {{- end }}
+{{/*
+Returns the list of controller-manager prometheus port names (one per line), matching the
+port names emitted by "spire-controller-manager.container" above. Used by the PodMonitor so
+its scrape endpoints stay in sync with the actual container ports. The suffix logic here MUST
+mirror the "spire-controller-manager.containers" iteration and its portSuffix computation.
+Emits nothing when the controller manager is disabled or prometheus telemetry is off.
+*/}}
+{{- define "spire-controller-manager.prometheus-port-names" }}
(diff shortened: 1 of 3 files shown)


kfox1111:
Sorry for the delay on this... I got a little confused when there was two prs in flight talking about doing the same thing. https://github.com/spiffe/helm-charts-hardened/pull/932 just merged. Does it fix all the issues you hit with this proposed pr? Should we close this one?

Munken:
> Sorry for the delay on this... I got a little confused when there was two prs in flight talking about doing the same thing. #932 just merged. Does it fix all the issues you hit with this proposed pr? Should we close this one?

No worries. Funny coincidence. 

#932 looks superior to this PR (common helpers to avoid drift). I’ll close this.

PR #932, opened 2026-08-26
Fix spire-server PodMonitor controller-manager port names

sabsari:
## What

The `spire-server` PodMonitor never scrapes controller-manager metrics: it targets a port named `prom-cm`, which no controller-manager container exposes. The main manager exposes `pm-cm`, and external managers expose auto-suffixed or overridden names.

## Changes

- Resolve controller-manager port names through shared helpers so the container ports and the PodMonitor endpoints can no longer drift.
- Enumerate every controller-manager that renders a prometheus port (main + external), honouring `prometheusPortName` overrides and the auto-suffix.
- Emit an endpoint only when its port renders; create no PodMonitor when none do.

## Notes

No chart version bump. Container rendering is byte-for-byte unchanged (pure refactor). Backward compatible.

### DIFF
--- a/charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
+++ b/charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
@@ -1,3 +1,72 @@
+{{/* 15-char-safe port-name suffix for a controller-manager cluster name ("" -> ""). Shared by container ports and the PodMonitor so they cannot drift. */}}
+{{- define "spire-controller-manager.portSuffix" -}}
+{{- $name := . -}}
+{{- $portSuffix := "" -}}
+{{- if ne $name "" -}}
+{{-   $portSuffix = printf "-%s" $name -}}
+{{-   if gt (len $name) 9 -}}
+{{-     $numberMatch := regexFind "[-]?[0-9]{1,2}$" $name -}}
+{{-     if $numberMatch -}}
+{{-       $numLen := len $numberMatch -}}
+{{-       $baseLen := sub (len $name) $numLen | int -}}
+{{-       $baseName := substr 0 $baseLen $name -}}
+{{-       if not (hasPrefix "-" $numberMatch) -}}
+{{-         $numberMatch = printf "-%s" $numberMatch -}}
+{{-       end -}}
+{{-       $maxBase := sub 9 (len $numberMatch) | int -}}
+{{-       $baseName = $baseName | trunc $maxBase | trimSuffix "-" -}}
+{{-       $portSuffix = printf "-%s%s" $baseName $numberMatch -}}
+{{-     else -}}
+{{-       $hash := sha256sum $name | trunc 3 -}}
+{{-       $portSuffix = printf "-%s-%s" ($name | trunc 5 | trimSuffix "-") $hash -}}
+{{-     end -}}
+{{-   end -}}
+{{- end -}}
+{{- $portSuffix -}}
+{{- end -}}
+
+{{/* Resolve a port name: override wins, else prefix+suffix. dict: prefix, portSuffix, override */}}
+{{- define "spire-controller-manager.portName" -}}
+{{- if and (hasKey . "override") (ne (.override | toString) "") -}}
+{{- .override -}}
+{{- else -}}
+{{- printf "%s%s" .prefix .portSuffix -}}
+{{- end -}}
+{{- end -}}
+
+{{/* Prometheus port name for one controller-manager. dict: name (""=main), settings (may hold prometheusPortName) */}}
+{{- define "spire-controller-manager.promPortName" -}}
+{{- $override := "" -}}
+{{- if hasKey .settings "prometheusPortName" -}}
+{{-   $override = .settings.prometheusPortName -}}
+{{- end -}}
+{{- include "spire-controller-manager.portName" (dict "prefix" "pm-cm" "portSuffix" (include "spire-controller-manager.portSuffix" .name) "override" $override) -}}
+{{- end -}}
+
+{{/* List of prometheus port names for every controller-manager that renders one. Consumed by the PodMonitor. */}}
+{{- define "spire-controller-manager.prometheusPortNames" -}}
+{{- $root := . -}}
+{{- $names := list -}}
+{{- if eq (.Values.controllerManager.enabled | toString) "true" -}}
+{{-   $names = append $names (include "spire-controller-manager.promPortName" (dict "name" "" "settings" .Values.controllerManager)) -}}
+{{- end -}}
+{{- if .Values.externalControllerManagers.enabled -}}
+{{-   $clusters := default .Values.kubeConfigs .Values.externalControllerManagers.clusters -}}
+{{-   range $name, $_ := $clusters -}}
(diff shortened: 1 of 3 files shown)


charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
```diff
@@ -1,3 +1,70 @@
...
+{{-   $override = .settings.prometheusPortName -}}
+{{- end -}}
+{{- include "spire-controller-manager.portName" (dict "prefix" "pm-cm" "portSuffix" (include "spire-controller-manager.portSuffix" .name) "override" $override) -}}
+{{- end -}}
+
+{{/* List of prometheus port names for every controller-manager that renders one. Consumed by the PodMonitor. */}}
+{{- define "spire-controller-manager.prometheusPortNames" -}}
+{{- $root := . -}}
+{{- $names := list -}}
+{{- if or (dig "telemetry" "prometheus" "enabled" .Values.telemetry.prometheus.enabled .Values.global) (and (dig "spire" "recommendations" "enabled" false .Values.global) (dig "spire" "recommendations" "prometheus" true .Values.global)) -}}
```
kfox1111: Any reason to gate here? Shouldn't the function return the list no matter if the telemetry is enabled or not?

charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
```diff
@@ -1,3 +1,70 @@
...
+{{-   if eq (.Values.controllerManager.enabled | toString) "true" -}}
+{{-     $names = append $names (include "spire-controller-manager.promPortName" (dict "name" "" "settings" .Values.controllerManager)) -}}
+{{-   end -}}
+{{-   if .Values.externalControllerManagers.enabled -}}
+{{-     $clusters := default .Values.kubeConfigs .Values.externalControllerManagers.clusters -}}
```
kfox1111: Could this produce duplicate entries? Should it be uniqued?

charts/spire/charts/spire-server/templates/podmonitor.yaml
```diff
@@ -2,6 +2,15 @@
 {{- if (dig "telemetry" "prometheus" "podMonitor" "enabled" .Values.telemetry.prometheus.podMonitor.enabled .Values.global) }}
 {{- $namespace := include "spire-server.podMonitor.namespace" . }}
 {{- $podNamespace := ( include "spire-server.namespace" . ) }}
+{{- $promEnabled := or (dig "telemetry" "prometheus" "enabled" .Values.telemetry.prometheus.enabled .Values.global) (and (dig "spire" "recommendations" "enabled" false .Values.global) (dig "spire" "recommendations" "prometheus" true .Values.global)) }}
```
kfox1111: I'm a little confused by line 5 here. why make a promEnabled flag when the whole object is gated by a conditional up at line 2.

charts/spire/charts/spire-server/templates/podmonitor.yaml
```diff
@@ -2,6 +2,15 @@
...
+{{- $endpoints := list }}
+{{- if $promEnabled }}
+{{-   $endpoints = append $endpoints "prom" }}
+{{- end }}
+{{- range (include "spire-controller-manager.prometheusPortNames" . | fromYamlArray) }}
+{{-   $endpoints = append $endpoints . }}
+{{- end }}
```
kfox1111: Might be simpler to just do the range once down at like  line 32, rather then build a list that is used once, then walk the list again to render the port list.

kfox1111 (commented):
Generally looks good. Some small issues inline.

charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
sabsari: The gate was there so the PodMonitor wouldn't list controller-manager ports that don't render when prometheus telemetry is off. But you're right; that's the caller's concern, not the resolver's. I moved the telemetry check up into the PodMonitor's top-level conditional.

charts/spire/charts/spire-server/templates/_controller-manager-container.tpl
sabsari: Added a check that fails if two controller-managers resolve to the same port name.

charts/spire/charts/spire-server/templates/podmonitor.yaml
sabsari: You're right, the separate flag was unnecessary. Folded the check into the top-level conditional.

charts/spire/charts/spire-server/templates/podmonitor.yaml
sabsari: Done. The port names are now ranged directly in `podMetricsEndpoints`.

charts/spire/charts/spire-server/templates/podmonitor.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if not .Values.externalServer }}
-{{- if (dig "telemetry" "prometheus" "podMonitor" "enabled" .Values.telemetry.prometheus.podMonitor.enabled .Values.global) }}
+{{- if and (dig "telemetry" "prometheus" "podMonitor" "enabled" .Values.telemetry.prometheus.podMonitor.enabled .Values.global) (or (dig "telemetry" "prometheus" "enabled" .Values.telemetry.prometheus.enabled .Values.global) (and (dig "spire" "recommendations" "enabled" false .Values.global) (dig "spire" "recommendations" "prometheus" true .Values.global))) }}
```
kfox1111: hmmm.. This changes behavior doesnt it? Not just, is podMonitors enabled, but if just prometheus is enabled. not all clusters with prometheus use or even support podMonitors. I think the original check is better?

tests/unit/spire_test.go
```diff
@@ -810,6 +810,70 @@ spire-server:
...
+  kubeConfigs:
+    child01:
+      kubeConfig: |
+        apiVersion: v1
+        kind: Config
+    verylongclustername:
+      kubeConfig: |
+        apiVersion: v1
+        kind: Config
+  externalControllerManagers:
```
kfox1111, line 857: Could you add another test with kubeConfigs used and externalControllerManagers.clusters set to  {}? This is the chart default and would be good to make sure works as expected with the default function being used.

charts/spire/charts/spire-server/templates/podmonitor.yaml
sabsari: The intent was to skip creating a no-op PodMonitor when prometheus is off, but I agree that's a bit too strict. Reverted line 2 to the original podMonitor.enabled check.

tests/unit/spire_test.go
sabsari: Added a test with `clusters: {}` and kubeConfigs set, covering the default fallback that derives the controller-managers from kubeConfigs.

kfox1111 (approved):
LGTM. Thanks! :)

sabsari:
Thanks for the detailed review, @kfox1111!

PR #936, opened 2026-08-30
Update spire-ha-agent

kfox1111:
Bumps version, adds its new broker api support, and metrics


charts/spire-ha-agent/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: spire-ha-agent
 description: A Helm chart to install the SPIRE HA agent.
 type: application
 version: 0.3.1
-appVersion: "0.3.0"
+appVersion: "0.5.0"
```
faisal-memon, line 6: The chart version still remains `0.3.1` while `appVersion` moves to `0.5.0`. Since this is an existing published chart version, the release process cannot publish these new templates for users to install. Could we bump `version` as well and regenerate the chart README?

charts/spire/charts/spire-agent/templates/roles.yaml
```diff
@@ -24,3 +34,48 @@ roleRef:
...
+
+{{- if $enforced }}
+{{- $trustDomain := include "spire-lib.trust-domain" . }}
+{{- range $key, $value := .Values.workloadAttestors.k8s.brokerAPI.brokers }}
+{{- /* Defaulting to false, not auto, is deliberate: a broker someone adds
+       themselves gets no grant unless they ask for one. */}}
+{{-   $imp := dig "impersonation" "clusterWidePodsOnly" false $value | toString }}
+{{-   $grant := ternary $enforced (eq $imp "true") (eq $imp "auto") }}
+{{-   if and (or (not (hasKey $value "enabled")) $value.enabled) $grant }}
+{{-     $idTemplate := (index $.Values.brokerAPI.brokers $key).idTemplate }}
```
faisal-memon, line 46: This assumes every enabled `workloadAttestors.k8s.brokerAPI.brokers` entry has a matching `brokerAPI.brokers` entry. If a custom broker is enabled only in the workload-attestor configuration, the new `auto` path resolves to `enforced` but this dereferences a missing map value and Helm fails to render. Could we use `dig`/a safe fallback here, or validate that the matching broker configuration is present?

faisal-memon (approved):

charts/spire/charts/spire-agent/templates/roles.yaml
kfox1111: This is true, but also by design. Every workloadattestor reference does need a matching brokerAPI.brokers entry. and every broker needs to be explicit about if they support passive or not. So, while burdonsom this is the best we can do right now. There is discussion on the spire side about what might be done to make this configuration surface better.

charts/spire-ha-agent/Chart.yaml
kfox1111: we bump it at release time

PR #939, opened 2026-09-01
Support Gateway API spec.infrastructure on the shared Gateway

sabsari:
## What

Add `gatewayAPI.gateway.infrastructure` to the shared Gateway, exposing the Gateway API standard `Gateway.spec.infrastructure` field (labels/annotations, GA since Gateway API v1.2).

## Changes

- `spire-lib`: render `spec.infrastructure` in the shared `gateway-resource` helper as a raw `toYaml` passthrough, guarded by `with` (same pattern as the existing `annotations` / `extraListeners`).
- `spire`, `spire-nested`: add the `gatewayAPI.gateway.infrastructure` value (default `{}`) and regenerate the READMEs.
- `tests/unit`: cover the passthrough render.

## Example

Inject cloud load-balancer annotations onto the provisioned Gateway Service — e.g. exposing it as an AWS internal NLB:

```yaml
gatewayAPI:
  gateway:
    enabled: true
    className: istio
    infrastructure:
      annotations:
        service.beta.kubernetes.io/aws-load-balancer-type: external
        service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
        service.beta.kubernetes.io/aws-load-balancer-scheme: internal
```

Verified with istio: the annotations/labels land on the Service istio auto-provisions for the Gateway.

## Notes

- Purely additive and backward compatible: the `with` guard means the default `{}` renders nothing, so existing output is byte-for-byte unchanged.


kfox1111 (approved):
LGTM. Thanks! :)

Issue #341, opened 2024-04-30
move spire-controller-manager to a separate pod


drewwells:
For background (skip if you know this), ingress and k8s services only send traffic to pods marked ready. If any container in the pod is not marked ready, no traffic will be sent to the pod. This is to handle zero downtime rotations of pods in replicasets. 

The spire-server and spire-controller-manager have different roles in spire. spire-server is responsible for API and serving requests. If it's down, especially in the statefulset deployment, spire eventually stops working entirely. However, spire-controller-manager is responsible for managing CRs in the cluster. If it's down, the impact is more nuanced.

Since these two containers are stuck in the same pod, when either of them are down, spire API is down. The controller needs to be moved to a separate pod so its outages do not impact spire itself. I'm facing a problem where a federated endpoint lost SSL cert. Controller Manager is restarting causing outages to spire (not just federation problems). 

```
2024-04-30T12:59:33Z    ERROR   setup   problem running manager {"error": "failed to wait for clusterspiffeid caches to sync: timed out waiting for cache to be synced for Kind *v1alpha1.ClusterSPIFFEID"}
main.run
        /workspace/main.go:347
main.main
        /workspace/main.go:82
runtime.main
        /usr/local/go/src/runtime/proc.go:250
2024-04-30T12:59:33Z    DEBUG   events  spire-server-0_84b56424-9ab4-495f-bf58-c2efca64d303 stopped leading     {"type": "Normal", "object": {"kind":"Lease","namespace":"spire-server","name":"8aa27f40.spiffe.io","uid":"5deab3b1-5f1d-4855-8cc5-f15bcdcbbee0","apiVersion":"coordination.k8s.io/v1","resourceVersion":"1272111663"}, "reason": "LeaderElection"}
2024-04-30T12:59:33Z    ERROR   error received after stop sequence was engaged  {"error": "leader election lost"}
sigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).engageStopProcedure.func1
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.1/pkg/manager/internal.go:555
```

faisal-memon:
Created an issue on the controller manger to see if there is interest in supporting this deployment mode. https://github.com/spiffe/spire-controller-manager/issues/363

PR #941, opened 2026-09-01
Stand alone controller manager

Munken:
## What

Experimental alternative to the bootstrap-sidecar/CR design in #938, for
granting the standalone `spire-controller-manager` Deployment (see #341)
its own SPIFFE identity without depending on the per-node
spiffe-csi-driver/spire-agent DaemonSet or any controller-manager
reconciliation loop for the bootstrap identity itself.

Approach: a "node alias" registration entry (`parentID =
spiffe://<td>/spire/server`, matched via `k8s_psat` node-attestor
selectors) plus an admin workload entry parenting the standalone
Deployment's own SPIFFE ID - both fully static/computable at Helm
render time - applied via a `postStart` hook on the `spire-server`
container's own admin socket, instead of a dedicated bootstrap
controller-manager/CR.

## Why

Explores whether the node-alias + static-entries pattern (also used by
the separate, upstream `spire-tcp-broker` chart for a different
purpose) can avoid the extra bootstrap ClusterSPIFFEID/reconciler
machinery in #938 altogether.

## Status

Was previously **known broken** (see 09e0549b): the `postStart` hook
raced spire-server's own startup and kubelet would permanently
CrashLoopBackOff the container.

This is now fixed and validated end-to-end in kind (see 096a4b75):
- `postStart` retries in a loop (via a busybox binary copied into a
  shared `emptyDir` by a new init container) until the admin socket is
  open.
- The node-alias SPIFFE ID moved off the reserved `/spire/agent/...`
  namespace to `/agent-alias/...`.
- The co-located `spire-controller-manager` container is attested via
  the `unix` WorkloadAttestor (matched by a fixed `unix:uid` selector)
  instead of the `k8s` WorkloadAttestor, avoiding a kubelet-reachability
  dependency that would otherwise require `hostNetwork: true`. Same
  pattern already used by `charts/spire-identity-exchange` in this repo.
  Credit to a contributor's feedback for pointing at the `unix`
  attestor instead of assuming `k8s` was required here.

Result: `spire-server-0` stays `Ready` with 0 restarts, and the
standalone controller-manager Deployment reaches `2/2 Running`, obtains
its own SVID, and successfully reconciles `ClusterSPIFFEIDs`.

## jwtSVIDExec kubeConfigs entries in standalone mode (8ca0ba4f)

Previously the chart failed at render time when a `kubeConfigs` entry
using `jwtSVIDExec` was combined with `controllerManager.deploymentMode
standalone`, because the exec plugin's default source (spire-server's
admin API socket) is not reachable from a separate standalone
Deployment's Pod.

The standalone Pod already carries its own in-pod `spire-agent` sidecar
and its own admin-scoped SPIFFE identity, so this now switches the exec
plugin's kubeconfig to `workload-api` mode against that Pod's own agent
socket and materializes `jwtSVIDExecConfig.spiffeID` as an extra static
bootstrap workload entry (non-admin, `unix:path` on the plugin binary,
tagged with a deterministic hint the kubeconfig references via
`SPIFFE_JWT_HINT`) parented to the same node-alias used by the
standalone Pod's own identity. No dependency on spire-server's admin
socket remains for the standalone consumer path.

Because a single Secret entry can only carry one kubeconfig, an entry
that would be consumed by both a standalone controller-manager
(workload-api mode) and an enabled spire-server-container consumer such
as `nodeAttestor.externalK8sPSAT` / `notifier.externalK8sBundle` /
`bundlePublisher.externalK8sConfigMap` (server-admin-api mode) is
rejected at render time with a clear message pointing at both
alternatives (split into two entries, or narrow the server-container
consumer's `clusters` map).

Unit tests cover the new positive path and the dual-consumer rejection.
The `standalone-controller-manager` example gains a
`values-jwt-svid-exec.yaml` overlay and a helm-template smoke assertion
in `run-tests.sh`.

## Not proposing to merge this as-is

Opening as a **draft**, for reference and discussion alongside #938,
not requesting review/merge. In particular still open:
- Whether this is preferable to #938's bootstrap-sidecar/CR approach,
  or whether the two should be reconciled into one design.
- The `postStart` hook + busybox-retry-loop mechanism is more fragile
  than a real reconciliation loop (e.g. no automatic recovery if the
  ConfigMap/ClusterSPIFFEID content changes after the Pod is created,
  relies on kubelet's postStart semantics).

## How this was tested

- `cd tests/unit && ginkgo` (70/70 passing)
- `helm lint` on both default (sidecar) and standalone
  (`spire-server.controllerManager.deploymentMode=standalone`) render
  paths, plus standalone + `jwtSVIDExec` render
- Manual end-to-end validation on a local kind cluster: fresh install,
  `spire-server-0` Pod readiness/restart count, `spire-controller-manager`
  container status, `spire-server entry show` to confirm the node-alias
  and admin workload entries and a subsequently-reconciled
  `ClusterSPIFFEID` entry.

Related: #341, #938


Munken:
Pushed a follow-up fix (8d24104a): the previous commit's `unix:uid:1000` selector hardcoded `podSecurityContext.runAsUser: 1000`, which breaks under OpenShift's default `restricted`/`restricted-v2` SCC (each namespace gets a random, SCC-assigned UID range; a hardcoded `runAsUser` outside it is rejected, and this repo never grants the more permissive `anyuid` SCC anywhere else).

Switched to a `unix:path` selector instead (`/spire-controller-manager`, the container's own binary path per its Dockerfile - same selector shape already used elsewhere in this chart's `values.yaml` for `spire-identity-exchange`). This doesn't need a known-ahead-of-time UID at all, so `podSecurityContext.runAsUser` now follows the same OpenShift-aware pattern already established by `spire-lib.podsecuritycontext`/`spire-identity-exchange.podSecurityContext`: forced to `1000` on vanilla k8s, left unset on OpenShift (`global.openshift: true`) so its SCC can assign the namespace's random UID.

Added a unit test asserting `runAsUser` is omitted under `global.openshift: true`, and re-validated end-to-end in kind: `spire-controller-manager-standalone` still reaches `2/2 Running`, obtains its SVID via the `unix:path` selector, and reconciles `ClusterSPIFFEIDs`. (Don't have an OpenShift/CRC cluster handy to validate the SCC interaction directly - only validated the Helm-render side of the OpenShift branch.)

Munken:
Pushed a follow-up feature (0693e26d): extended `controllerManager.deploymentMode: standalone` to support the existing `externalControllerManagers`/`kubeConfigs` mechanism (already used by sidecar mode to reconcile external/downstream clusters).

For each external cluster, this adds a **fully separate standalone Deployment** (its own in-pod `spire-agent` sidecar, own SPIFFE identity/bootstrap entries, own ServiceAccount) alongside the existing default/local instance - rather than N sibling containers sharing one Pod like sidecar mode does - so a problem with one cluster's kubeconfig or reconciliation can't affect any other instance's Pod.

Along the way this surfaced and fixed two real bugs in the existing standalone-mode plumbing:
- The postStart hook's retry loop treated `AlreadyExists` (which the `spire-server` CLI reports as a nonzero-exit error) as a failure, so it would retry 30 times and then permanently fail - and kubelet would kill the container - on every restart/upgrade once the bootstrap entries already existed from a previous run, rather than only on a genuine failure.
- Enabling `externalControllerManagers` under `deploymentMode: standalone` was *also* adding non-functional, wrongly-addressed sidecar containers to the spire-server Pod itself (sidecar mode's own external-cluster container-generation loop fired unconditionally, regardless of `deploymentMode`).

Validated end-to-end in kind: with two `kubeConfigs` entries both pointing at the same kind cluster (simulating two remote clusters without needing a second real one), `spire-server-0` stays `1/1 Running` with 0 restarts, and all three standalone Deployments (default + 2 external) reach `2/2 Running`, each obtaining its own SVID and successfully reconciling its own `ClusterSPIFFEID` against its respective (simulated) remote cluster. 68/68 unit tests pass.

Known limitation: `jwtSVIDExec`-based `kubeConfigs` entries aren't supported in standalone mode (that mechanism mints its exec-credential JWT-SVID via spire-server's own admin socket, which a separate standalone Deployment's Pod doesn't have access to) - combining the two now fails at render time with a clear error message.

Munken:
> Known limitation: `jwtSVIDExec`-based `kubeConfigs` entries aren't supported in standalone mode (that mechanism mints its exec-credential JWT-SVID via spire-server's own admin socket, which a separate standalone Deployment's Pod doesn't have access to) - combining the two now fails at render time with a clear error message.

@kfox1111 have indicated that this should work with controller manager acquiring a SVID via the Workload API

Munken:
The known limitation around `jwtSVIDExec`-based `kubeConfigs` entries in standalone mode is now resolved in 8ca0ba4f.

Rather than the exec plugin sourcing its JWT-SVID from spire-server's admin socket (which the separate standalone Deployment's Pod cannot reach), the plugin now sources it from the standalone Pod's own in-pod `spire-agent` sidecar's Workload API socket. `jwtSVIDExecConfig.spiffeID` is materialized as an extra static bootstrap workload entry (non-admin, matched via `unix:path` on the plugin binary, tagged with a deterministic hint the kubeconfig references via `SPIFFE_JWT_HINT`) parented to the same node-alias used by the standalone Pod's own identity, so the user-facing `values.yaml` contract for `jwtSVIDExecConfig.spiffeID` is preserved.

The old blanket render-time `fail` is replaced with a narrower one that only fires when a single entry would be consumed by both a standalone controller-manager (workload-api mode) and an enabled spire-server-container consumer such as `nodeAttestor.externalK8sPSAT` / `notifier.externalK8sBundle` / `bundlePublisher.externalK8sConfigMap` (server-admin-api mode), since one Secret entry can only carry one kubeconfig.

Covered by new unit tests (positive path + dual-consumer rejection) and a `helm template` smoke assertion in the standalone-controller-manager example.

PR #945, opened 2026-09-04
Revert "Bump k8s.io/apimachinery from 0.36.2 to 0.37.0 in /tests"

kfox1111:
Reverts spiffe/helm-charts-hardened#940


kfox1111:
These tests cant pass since there was two changes that happened back to back that broke things. one has to be merged before all tests go back to green.

kfox1111:
Tests cant pass on this since two bad patches were merged and both have to be reverted before everythings fixed.

PR #953, opened 2026-09-14
Fix cacheNamespaces wiring for the local and external controller mana…

### LINKED ISSUE #668: spire-server controller manager configmap cacheNamespaces wouldnt render
Hello, I could not get this [line](https://github.com/spiffe/helm-charts-hardened/blob/3c3718c904da320157ac6536999fc3cf56177013/charts/spire/charts/spire-server/templates/controller-manager-configmap.yaml#L58) to render on the resulting configmap to limit the scope of the manager's cache.

Other occurrences of a similar check uses a hasKey (e.g. `{{- if hasKey .settings "ignoreNamespaces" }}`), I'm wondering if the check for cacheNamespaces should be the same? 

My pertinent values config (in jsonnet) is:
```
controllerManager: {
  enabled: true,
  cacheNamespaces: {
    nsWithLabel: {
      labelSelectors: {
        "namespace-type": "app",
      },
    },
  },
},
```

Thank you!

JoelGoh92:
## Summary

 `cacheNamespaces` never rendered into the `ControllerManagerConfig` CR for two of the three places you'd set it:
  - the single/primary controller manager's own `cacheNamespaces`
  - `externalControllerManagers.defaults.cacheNamespaces` (apply to every
    dynamically-generated child-cluster controller manager)

  Only `externalControllerManagers.clusters.<name>.cacheNamespaces` (the per-cluster override case) actually worked.

  ## Root cause
  Unlike every other settings-aware field in this template (`filterByClassName`, `watchClassless`, `ignoreNamespaces`, etc.), `cacheNamespaces` only ever read `.settings.cacheNamespaces` with no fallback to `.defaults.cacheNamespaces`.
  For the primary controller manager, `.settings` is always an empty stub, so the field was structurally unreachable regardless of what was set in values.

  Fixed by adding the same `hasKey .settings "cacheNamespaces" ... else .defaults.cacheNamespaces` fallback 

  ## Test plan
  - Added 3 Ginkgo specs under `spire-server.controllerManager.cacheNamespaces`
    covering: primary CM, `externalControllerManagers.defaults`, and per-cluster
    override precedence.
  - Verified the new specs fail against the pre-fix template and pass after the
    fix (regression-proven, not just a happy-path assertion).
  - Full existing unit suite (55 specs) passes unchanged.

  Fixes #668


kfox1111 (approved):
LGTM. Thanks! :)

PR #954, opened 2026-09-14
spiffe-csi-driver: set atuomountServiceAccountToken to false

daescha:
spiffe csi driver doesnt need the token, no need to mount it.


kfox1111 (approved):
LGTM. Thanks! :)

PR #955, opened 2026-09-14
fix(spiffe-csi-driver): harden registrar context

daescha:
securityContext now settable for node registrar context, and it can take the recommended defaults like the others can, with the exception of run as non root.


charts/spire/charts/spiffe-csi-driver/values.yaml
```diff
@@ -143,6 +143,10 @@ nodeDriverRegistrar:
     tag: v2.15.0
   ## @param nodeDriverRegistrar.extraEnvVars [array] Extra environment variables to be added to the nodeDriverRegistrar container
   extraEnvVars: []
+  ## @param nodeDriverRegistrar.securityContext [object] Security context for the nodeDriverRegistrar container, merged over the recommended defaults
+  ##
+  securityContext:
+    runAsNonRoot: false
```
kfox1111: Not sure what this setting does vs not having it here at all which is chart standard.

Also, nothing is testing that setting this value to a non root value actually works. Does it? Its got to be able to talk to kubelet I think and that may require root anyway?

marcofranssen: I propose that value runAsNonRoot is hardcoded in the template and ignored when reading the values.

Then other settings can be given by enduser without them be able to run it as non root

daescha: okay ill make it hardcoded :+1:

daescha: made hardcoded and confirmed that running with anything other than a root user would fail: the default image has root user, so that fails with container config error, but an altered image would also fail because `/var/lib/kubelet/plugins_registry` that the registrar container writes to requires root.

tests/unit/spire_test.go
```diff
@@ -424,6 +424,38 @@ spire-server:
...
+			Expect(registrar).Should(ContainSubstring("readOnlyRootFilesystem: true"))
+			Expect(registrar).Should(ContainSubstring("type: RuntimeDefault"))
+			Expect(registrar).ShouldNot(ContainSubstring("privileged: true"))
+		})
+		It("takes other settings from values but ignores runAsNonRoot", func() {
+			objs, err := ValueStringRender(chart, `
+spiffe-csi-driver:
+  nodeDriverRegistrar:
+    securityContext:
+      runAsUser: 1234
```
kfox1111, line 488: still wondering if this test makes sense. would helm install and then fail in cluster in this config?
shoudld we force the runAsUser to be there or empty but the value has to be 0?

daescha: made the test to assert its forced to 0 user and run as nonroot: false

kfox1111 (approved):
LGTM. Thanks! :)

PR #956, opened 2026-09-14
fix(spiffe-csi-driver): make  validatingAdmissionPolicy usable when plugin name is altered 

daescha:
cant use it currently because we alter the plugin name to allow installation of multiple deployments including spire charts to be installed to the same cluster.


kfox1111 (approved):
LGTM. Thanks! :)

PR #957, opened 2026-09-14
spire-agent: narrow the OpenShift SCC

daescha:
- hostipc never used
- hostports only possibly used if user adds one in the extra init/containers
- privileged is only needed if tpmdirect is used or user overrode a securitycontext
- allowprivilegeescalation is only needed if user overrode a security context


charts/spire/charts/spire-agent/templates/scc-spire-agent.yaml
```diff
@@ -20,12 +45,12 @@ volumes:
   - emptyDir
 allowedCapabilities: null
 allowHostDirVolumePlugin: true
-allowHostIPC: true
```
kfox1111, line 23: This isnt used for socket communication? like, workload api, delegated api, or broker api?

daescha: No, those APIs are all Unix-domain sockets exposed through the hostPath socket mounts (`spire-agent-socket-dir`, `-admin-socket-dir`, `-broker-socket-dir`), which is what `allowHostDirVolumePlugin: true` covers. `hostIPC` is a different thing: it puts the pod in the host's System V IPC namespace (shared memory segments, semaphores and message queues) which nothing in the agent uses. The only host namespace the agent needs is `hostPID` (the k8s workload attestor resolves caller PIDs against the host). So `allowHostIPC: false` is safe.

kfox1111 (approved):
LGTM. Thanks! :)

PR #958, opened 2026-09-16
feat(spire-server): add keyManager.awsKMS.enableTagBasedKeyDiscovery

bpalermo:
## What

Adds `keyManager.awsKMS.enableTagBasedKeyDiscovery` (default `false`) to the spire-server chart. When set, the `aws_kms` KeyManager gets `enable_tag_based_key_discovery: true` in its `plugin_data`.

## Why

SPIRE 1.15.2 added tag-based key discovery to the `aws_kms` plugin (spiffe/spire#7006): keys are discovered through the AWS Resource Groups Tagging API using the tags SPIRE applies to them, instead of listing every key and alias in the account. Alias-based discovery, still the plugin default, is deprecated upstream and SPIRE 1.15.3 logs a warning about it at every server start; it will be removed in a future release. Until now the only way to set the flag through the chart was `unsupportedBuiltInPlugins`.

## Changes

- `values.yaml`: new `keyManager.awsKMS.enableTagBasedKeyDiscovery` next to `keyTags`, with the IAM requirements in the description (`kms:TagResource`, `tag:GetResources`; the latter is identity-policy only and cannot be granted through a KMS key policy).
- `templates/configmap.yaml`: renders the flag only when true, so existing installs are byte-for-byte unchanged.
- `README.md`: regenerated with `./helm-docs.sh` (one row added).
- `tests/unit/spire_test.go`: asserts the key is absent by default and present when enabled.
- `examples/aws-kms-tag-discovery/`: values + README covering the automatic alias→tag migration and the IAM delta, mirroring `examples/aws-kms-tags`.

No chart version bump, per the contribution guidelines.

## Testing

- `cd tests/unit && go test ./...` passes (including the two new assertions).
- `helm template` with the example values renders `"enable_tag_based_key_discovery": true` inside `aws_kms.plugin_data`; without the value the rendered `server.conf` is unchanged.
- Running in production on SPIRE 1.15.3 / chart 0.30.2 (via `unsupportedBuiltInPlugins` with the identical rendered block): the plugin tagged the existing alias-managed keys on first start and discovery has been tag-based since.


kfox1111 (approved):
LGTM. Thanks! :)

